PR #8, opened 2017-09-29
Cache gene distributions per experiment when start up the app

elisabetb:
Code has a script that extracts from all baseline experiments the experimentAccession and the experimentType and uses them to access the json URL to warm the gene distribution for each of them.


wbazant:
@elisabetb I fixed this now, feel free to review and if you think my changes are okay then merge

PR #11, opened 2017-10-04
Task 150785439

wbazant:


base/src/main/java/uk/ac/ebi/atlas/bioentity/properties/BioEntityPropertyService.java
```diff
@@ -98,7 +97,7 @@ public BioEntityPropertyService(UniProtClient uniProtClient,
                 .collect(Collectors.toList());
     }
 
-    private JsonArray bioentityProperties(String identifier, Species species,
+    JsonArray bioentityProperties(String identifier, Species species,
```
alfonsomunozpomer: In principle I don’t think it’s good practice to expose methods just for testing. Only public methods should be tested unless, in doing so, you need a lot of work to set the test up.

In this case you can test pretty much the same cases with the `modelAttributes` method and passing an `entityName`.

base/src/test/java/uk/ac/ebi/atlas/bioentity/properties/BioEntityPropertyServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class BioEntityPropertyServiceTest {
+
+
+    @Mock
+    UniProtClient uniProtClient;
```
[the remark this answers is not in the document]
wbazant: Thanks, fixed!

base/src/main/java/uk/ac/ebi/atlas/bioentity/properties/BioEntityPropertyService.java
wbazant: Well, it's an opinion! I disagree with it. A more pragmatic approach is to do whatever you need to make tests easier. The "private" being there or not doesn't give you anything, especially since when working in an IDE you can easily check where the method is called in your codebase.

It's a smell and shows that a public method is hard to test but it is the case. `Map<String, Object>` returned from `modelAttributes`. I could call that but then I would also need to cast the result etc. and there's no reason to it - no better code as a result, not making someone less likely to break the code later, the test code isn't better, it's just more effort.

alfonsomunozpomer: Well, the purpose of the class isn’t to be tested. :) It’s supposed to provide an API defined by its public methods. The rest are implementation details. But I also added the caveat that sometimes exposing a private method can simplify testing (test-driven design, I guess). But it’s not really the case here.

Compare this:
```
assertThat(
                subject.bioentityProperties(identifier, species, ImmutableList.of(), ImmutableMap.of()),
                is(new JsonArray())
        );
```

To this, no casting necessary:
```
assertThat(
                subject.modelAttributes(identifier, species, ImmutableList.of(), "foobar", ImmutableMap.of()).get("bioentityProperties"),
                is(gson.toJson(new JsonArray()))  // or just .is("[]")
        );
```

Remember: “Typing isn’t the bottleneck”. ;)

wbazant: You don't want to assert a JSON object is a specific string, because when you change such a test you need to make sure all the whitespace is okay. Try to convert the test `outputLooksRight`into how you propose it above, and compare with what's there now, to see what I mean.

wbazant: Also, it's a thought you might dislike but the purpose of the class isn't anything, unlike yours, or mine for that matter. It provides a functionality, it does it not well enough, and you need to make it do the right thing again: correctly, understandably, in a way such that another change won't probably break it, will make it easy to change in the future, and will help you solve other similar problems in the future. 

Acting on opinions about how to write code that don't help you accomplish the goal above (or a similar one) is about as good as trying to write it without using a letter E, or with using variable names that are just so barely marginally acceptable that you won't get told off for them: perfectly acceptable if you're working on something useless and don't actually want anything done. But here we're creating positive moral value with our work and (almost) carrying the world of science on our strained shoulders.

alfonsomunozpomer: I’m not completely sure on what your point is.

Rather than talking about semantics (e.g. if the purpose of a class isn’t anything, what’s the reason of its existence?), or analogies that fail short of illustrating a point or plainly a straw man (writing without using  the letter E), I’d rather talk about specific issues.

The reason of having code reviews is not only ensuring correctness but also having better, cleaner code. My opinion is that private methods shouldn’t be tested; qualifying something as an opinion doesn’t make it less valid. It’s as valid as the reasons backing it, and we can talk about them. However, and I pointed it out from the very beginning, it’s not a practice that should be followed blindly and there are cases where it’s a good idea not doing so. I’m more interested in discussing the latter.

You pointed out that comparing JSON isn’t good, and you’re right. I overlooked the `outputLooksRight` test. All in all, I have some issues with this class that fall out of the task at hand, though, and we can discuss them.

PR #14, opened 2017-10-05
Task 151517653

wbazant:
pr


base/src/main/java/uk/ac/ebi/atlas/search/analyticsindex/solr/AnalyticsSolrQueryTree.java
```diff
@@ -249,4 +241,18 @@ public static AnalyticsSolrQueryTree createForIdentifierSearch(SemanticQuery gen
...
+    Quotes differentiate between searching for a term or for a phrase.
+    https://lucene.apache.org/solr/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-SpecifyingTermsfortheStandardQueryParser
+
+    If the term has spaces inside, like "zinc finger" we don't want zinc OR finger, but two tokens : zinc and then finger.
+
+    If the query term has bits of syntax inside, quote it too.
+
+    Possibly we could quote everything, and I'm not sure why we're not.
+     */
+    static boolean searchValueNeedsQuotes(String searchValue){
```
alfonsomunozpomer: So, throughout all Solr and Lucene documentation, query terms are either words (without double quotes) or phrases with quotes. However, I haven’t really noticed any difference in behaviour when searching for a single term enclosed in double quotes. My impression is that double quotes (or their absence) give a clue to Solr to optimize a query.

Also, enclosing in quotes terms with colons is definitely from not-so-distant time when we (here I certainly speak for myself) knew very little about Solr. I think the right thing to do would be something like:
```
static String formatQueryTerm(String queryTerm) {
    String escapedQueryTerm = ClientUtils.escapeQueryChars(queryTerm);
    return escapedQueryTerm.contains(" ") ? StringUtils.wrap(escapedQueryTerm, "\"") : escapedQueryTerm;
}
```

gxa: The only, super-edge-case-very-improbable-scenario is that a very savvy user wants to use Solr wildcards or fuzzy search operator which the above wouldn’t allow for. We have every few days a couple of exceptions from users trying to use double quotes and logical operators in the search fields. I don’t think we should care about it, but we can discuss it.

alfonsomunozpomer: The only extremely-unlikely-edge-case-very-improbable-scenario I can think of is that a very savvy user wants to use Solr wildcards or the fuzzy search operator, which the above wouldn’t allow for. We have every few days a couple of exceptions from users trying to use double quotes and logical operators in the search fields. I don’t think we should care about it, but we can discuss it.

wbazant: The solution above is inadequate in the case where a query term accidentally contains bits of Solr syntax which turns the query into "something else". See the test case with that compound Laura gave.

gxa: I tested it and it works fine.

alfonsomunozpomer: I tested it and it works fine.

alfonsomunozpomer: Yeah, I was reading some weeks ago a couple of discussions in Drupal forums about a similar issue. It revolved about allowing some of Lucene’s special syntax or to add some custom wildcard characters to make searches more flexible. There wasn’t a consensus after some months because of the reasons you point out.

alfonsomunozpomer:
Build pass tests:
- http://bar:8085/browse/ATLAS-BASECIDEVTASKS7
- http://bar:8085/browse/ATLAS-GXACIDEVTASKS6

PR #16, opened 2017-10-11
Task 151043572

elisabetb:
This is the code for extracting comparisons - pathways and gene profiles.

Here you can find if the tests passed
http://bar.ebi.ac.uk:8085/browse/ATLAS-GXACIDEVTASKS0

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
@@ -93,8 +93,10 @@ public class DataFileHub {
         return new MicroarrayExperimentFiles(experimentAccession, arrayDesign);
     }
 
-    public DifferentialExperimentFiles getReactomePathwaysCFiles(String experimentAccession, String comparison) {
-        return new DifferentialExperimentFiles(experimentAccession, comparison);
+    public AtlasResource<TsvReader> getReactomePathwaysFiles(String experimentAccession, String comparison) {
+        return new TsvFile.ReadOnly(
+                dataFilesLocation, REACTOME_PATHWAYS_FILE_PATH_TEMPLATE,
+                experimentAccession, comparison);
     }
 
     public AtlasResource<KryoFile.Handle> getKryoFile(String experimentAccession,
@@ -205,7 +207,6 @@ public class DataFileHub {
 
     public class DifferentialExperimentFiles extends ExperimentFiles {
         public AtlasResource<ObjectInputStream<String[]>> percentileRanks;
-        public AtlasResource<TsvReader> reactomePathways;
 
         DifferentialExperimentFiles(String experimentAccession) {
             super(experimentAccession);
@@ -216,13 +217,6 @@ public class DataFileHub {
 
         }
 
-        DifferentialExperimentFiles(String experimentAccession, String comparison) {
-            this(experimentAccession);
-            this.reactomePathways =
-                    new TsvFile.ReadOnly(
-                            dataFilesLocation, REACTOME_PATHWAYS_FILE_PATH_TEMPLATE,
-                            experimentAccession, comparison);
-        }
     }
 
     public class RnaSeqDifferentialExperimentFiles extends DifferentialExperimentFiles {
--- a/base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
@@ -50,6 +50,8 @@ public class DataFileHub {
     final static String MICROARRAY_LOG_FOLD_CHANGES_FILE_PATH_TEMPLATE = "/magetab/{0}/{0}_{1}-log-fold-changes.tsv";
     final static String COEXPRESSION_FILE_TEMPLATE = "/magetab/{0}/{0}-coexpressions.tsv.gz";
 
+    final static String REACTOME_PATHWAYS_FILE_PATH_TEMPLATE = "/magetab/{0}/{0}.{1}.reactome.gsea.tsv";
+
     @Inject
     public DataFileHub(@Value("#{configuration['dataFilesLocation']}") String dataFilesLocation){
         Validate.notNull(
@@ -89,6 +91,10 @@ public class DataFileHub {
 
     public MicroarrayExperimentFiles getMicroarrayExperimentFiles(String experimentAccession, String arrayDesign) {
         return new MicroarrayExperimentFiles(experimentAccession, arrayDesign);
+    }
+
+    public DifferentialExperimentFiles getReactomePathwaysCFiles(String experimentAccession, String comparison) {
+        return new DifferentialExperimentFiles(experimentAccession, comparison);
     }
 
     public AtlasResource<KryoFile.Handle> getKryoFile(String experimentAccession,
@@ -198,13 +204,24 @@ public class DataFileHub {
     }
 
     public class DifferentialExperimentFiles extends ExperimentFiles {
-        public final AtlasResource<ObjectInputStream<String[]>> percentileRanks;
+        public AtlasResource<ObjectInputStream<String[]>> percentileRanks;
+        public AtlasResource<TsvReader> reactomePathways;
 
         DifferentialExperimentFiles(String experimentAccession) {
             super(experimentAccession);
             this.percentileRanks =
                     new TsvFile.ReadAsStream(
                             dataFilesLocation, DIFFERENTIAL_PERCENTILE_RANKS_FILE_PATH_TEMPLATE, experimentAccession);
+
+
+        }
+
+        DifferentialExperimentFiles(String experimentAccession, String comparison) {
+            this(experimentAccession);
+            this.reactomePathways =
+                    new TsvFile.ReadOnly(
+                            dataFilesLocation, REACTOME_PATHWAYS_FILE_PATH_TEMPLATE,
+                            experimentAccession, comparison);
         }
     }
 
@@ -221,6 +238,7 @@ public class DataFileHub {
                     new TsvFile.ReadOnly(
                             dataFilesLocation, DIFFERENTIAL_RAW_COUNTS_FILE_PATH_TEMPLATE, experimentAccession);
         }
+
     }
 
     public class MicroarrayExperimentFiles extends DifferentialExperimentFiles {
--- a/gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
+++ b/gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
@@ -81,12 +81,13 @@ public class DifferentialPathwaysComparisonService <Expr extends DifferentialExp
 
     public Map<Contrast, SetMultimap<String, Pair<String, DifferentialExpression>>> constructPathwaysByComparison (E experiment,
                                                                                                                    K preferences) {
-        //Read pathwaysId per comparison from file
+        //Read pathwaysId per comparison from file ({comparison}-reactome.gsea.tsv)
         Map<Contrast, List<String>> contrastPathwaysListMap = getPathwaysListPerComparison(experiment);
         SetMultimap<String, Pair<String, DifferentialExpression>> pathwayGeneExpressionMap = HashMultimap.create();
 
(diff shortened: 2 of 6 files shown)


base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
```diff
@@ -91,6 +93,10 @@ public MicroarrayExperimentFiles getMicroarrayExperimentFiles(String experimentA
         return new MicroarrayExperimentFiles(experimentAccession, arrayDesign);
     }
 
+    public DifferentialExperimentFiles getReactomePathwaysCFiles(String experimentAccession, String comparison) {
```
wbazant: This isn't needed because you can use get DifferentialExperimentFiles()

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
```diff
@@ -0,0 +1,133 @@
...
+
+        return comparisonsPathwaysMap;
+    }
+
+    private List<String> fetchPathwaysFromFile (String experimentAccession, Contrast comparison) {
+        Map<String, Double> result = new HashMap<>();
+
+        List<String[]> lines = dataFileHub.getReactomePathwaysCFiles(experimentAccession, comparison.getId()).reactomePathways.get().readAll();
+
+        for(int i = 1; i < lines.size(); i++) {
```
wbazant: Why not: 
```
for (String [] line: lines.subList(1, lines.size() -1)) {
...
}
```

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
```diff
@@ -0,0 +1,133 @@
...
+            String[] strings = lines.get(i);
+            String pathway = strings[0];
+            Double pvalue = Double.valueOf(strings[4]);
+
+            if (pvalue < 0.05) {
+                result.put(pathway, pvalue);
+            }
+        }
+
+        Map<String, Double> ordered = result.entrySet()
```
wbazant: You could do this here:
```
return result.entrySet().stream().sorted(Map.Entry.comparingByValue()).limit(10).map(Map.Entry::getKey).collect(Collectors.toList())
```

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+    private AssayGroup referenceAssay1 = new AssayGroup("g1", "assay1");
+    private AssayGroup testAssay1 = new AssayGroup("g2", "assay2");
+    private AssayGroup referenceAssay2 = new AssayGroup("g3", "assay3");
+    private AssayGroup testAssay2 = new AssayGroup("g4", "assay41", "assay42");
+
+    private Contrast c1 = new Contrast("g3_g1", null, referenceAssay1, testAssay1, "first contrast");
+    private Contrast c2 = new Contrast("g3_g2", null, referenceAssay2, testAssay2, "second contrast");
+
+    private String[] c1s1 = new String[]{"Term", "Accession", "Genes (tot)", "Stat (non-dir.) p", "p adj (non-dir.)"};
```
wbazant: This is hard to read. Doesn't matter now, but you could have also:
List<String[]> file1 = ImmutableList.of(
new String[]{"R-MMU-109581", "Apoptosis", "45", "0.535169358108118", "1"},
new String[]{"R-MMU-109581", "Apoptosis", "45", "0.535169358108118", "2"},
...
);

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        when(mockDifferentialExperiment.getAccession()).thenReturn(accession);
+        when(mockConfigurationTrader.getMicroarrayExperimentConfiguration(accession)).thenReturn(mockMicroarrayExperimentConfiguration);
+        when(mockConfigurationTrader.getMicroarrayExperimentConfiguration(accession).getContrasts()).thenReturn(Lists.newArrayList(c1,c2));
+
+        setUpDataForTest();
+    }
+
+    @Test
+    public void testGetPathwaysListPerComparison() throws Exception {
```
wbazant: This is a okay test - it will tell us when someone makes a change to the code and it'll be wrong - but it won't be easy to work with when something is failing. A better test explains why the code does what it does, and asserts various things about its behaviour.
E.g.
"only values with pValue less than 0.05 are taken into account".
"if there is no file, throw an exception"
"if there are many pathways to choose from return top ten pathways"

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+        assertThat(multimap.size(), is(1));
+
+        //Contrast 2
+        HashMultimap values = (HashMultimap) multimap.get(c2);
+        Map map = values.asMap();
+
+        Set<String> expectedPathways = Sets.newHashSet("R-MMU-416482", "R-MMU-397014", "R-MMU-199418", "R-MMU-193648", "R-MMU-1483206");
+        assertThat(map.keySet(), is(expectedPathways));
+
+        Collection<Pair<String, DifferentialExpression>> pairCollection1 = (Collection<Pair<String, DifferentialExpression>>) map.get("R-MMU-416482");
```
wbazant: This looks like it took lots of effort, and maybe it wasn't required.
For these kinds of tests I often do
```
assertThat(result, is(<very big object>))
```

When the result is JSON, I type in this first:
```
assertThat(result, is(new Gson().fromJson("{}", JsonObject.class)));
```
This of course fails, because the result is not "{}", and the failing test tells me what it was instead. Then I take what the result actually was and paste it in. :)

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
```diff
@@ -0,0 +1,133 @@
...
+        //For each comparison and each pathway get differential profiles (geneIds and differential expressions)
+        for (Map.Entry<Contrast, List<String>> comparison : contrastPathwaysListMap.entrySet()) {
+            for (String pathwayId : comparison.getValue()) {
+                preferences.setGeneQuery(SemanticQuery.create(pathwayId));
+                R requestContext = differentialRequestContextFactory.create(experiment, preferences);
+
+                DifferentialProfilesList<P> differentialProfiles = profilesHeatMap.fetch(requestContext);
+
+                if (!differentialProfiles.isEmpty()) {
+                    for (int i = 0; i < differentialProfiles.getTotalResultCount(); i++) {
```
wbazant: I think this is a bug, and that your code will sometimes break here.

differentialProfiles.getTotalResultCount() can be more than the amount of elements in the list. We use it to display on the experiment page something like "Showing 50 of 9871 genes found".

Please use a foreach loop, and I would suggest it to be the default practice when you don't use the index for anything:
```
for(P profile: differentialProfiles){
//
}
```

elisabetb: Ok, you are right. I didn't think that this can be higher to thousands, so I will change it.

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
elisabetb: Yes sure, I could have done it that way. 
I suppose this is a matter of programming styles, but there is nothing wrong with creating the strings separately and later add them in the empty list

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
elisabetb: I can add a failing test when there is no file or there is something wrong

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
elisabetb: Actually not because I have a second constructor for DifferentialExperimentFiles with an extra parameter and that is the one I need, not the getDifferentialExperimentFiles

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
elisabetb: I got it from the java 8 documentation for sorting the entrySet, and don't see anything wrong there

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
elisabetb: Well, actually I did it this way because I got one case that the expected result was different and needed to be sure that the code was returning the "expectedPathways". I know it has lot of effort and that is why it took me a while

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
wbazant: Ah right I didn't notice the new constructor. Instead of that constructor how about adding a method to DifferentialExperimentFiles?
```
public AtlasResource<TsvReader> getReactomePathways(String comparison){
   return new TsvFile.ReadOnly(
                             dataFilesLocation, REACTOME_PATHWAYS_FILE_PATH_TEMPLATE,
                             experimentAccession, comparison);
}
```

Adding a constructor that changes what the class is for - it used to represent one experiment, now it represents experiment and a comparison? - isn't good OO (it's not just a matter of programming style), and we can do better by using the above. You can see there was a similar need in e.g. BaselineExperimentFiles that needed a different data file depending on the unit, and that was a solution.

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
wbazant: You're making a new LinkedHashMap (with no reason that it should be precisely the LinkedHashMap), and then you're taking the keys of that hashmap, throwing away the values. The above is doing the same thing without the funky unnecessary part.

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonServiceTest.java
wbazant: Yea like I said you've created lots of work for yourself.

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
elisabetb: Good, I am making that change, it didn't come into my mind at that time. but just to clarify "overloaded constructor" is part of the OO programming, so is not a bad practice as you have referred to.

wbazant: (Y)! Yes sometimes multiple constructors are a good idea. Don't use them to change what the class represents when you only need to add one use case, that's the bad practice part.

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/differential/DifferentialPathwaysComparisonService.java
elisabetb: Ok done! I made the last change you proposed

wbazant:
@alfonsomunozpomer do approve / merge

elisabetb:
[Delivers #151043572]

PR #20, opened 2017-10-16
Task 152003874

alfonsomunozpomer:
There’s a timeout error if we query Reactome API using HTTPS, it works fine over HTTP. It works well on my laptop, so it’s an environment issue (it fails also in `lime`), so I’m inclined to think it has something to do with the EBI proxy servers.

In any case, if something wrong happens in `ReactomeClient`, would you rather return an empty string or something along the lines of *There was an error accessing Reactome, please try again later*? The problem with the latter is that it could give the impression that Reactome is failing when it is not... anyway, suggestions welcome!

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/bioentity/geneset/GeneSetPropertyService.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/bioentity/geneset/GeneSetPropertyService.java
@@ -52,7 +52,7 @@ public class GeneSetPropertyService {
 
     private Map<BioentityPropertyName, Set<String>> propertyValuesByType(BioentityPropertyName which,
                                                                          String identifier, String value) {
-        return ImmutableMap.of(which, (Set<String>)ImmutableSet.of(identifier), DESCRIPTION, ImmutableSet.of(value));
+        return ImmutableMap.of(which, ImmutableSet.of(identifier), DESCRIPTION, ImmutableSet.of(value));
     }
 
 }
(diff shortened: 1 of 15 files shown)


base/src/main/java/uk/ac/ebi/atlas/utils/ReactomeClient.java
```diff
@@ -32,7 +32,7 @@ public String fetchPathwayNameFailSafe(String reactomeId) {
             return StringUtils.trim(restTemplate.getForObject(url, String.class));
         } catch (RestClientException e) {
             LOGGER.warn("Reactome ID \"{}\" could not be found", reactomeId);
-            return null;
```
wbazant: This is called in a method `getLinkTextOrNull`, which is called in `Optional.ofNullable(...).orElse(<a decent default>)`. If you return "" are you not stopping the default from being displayed?

base/src/main/java/uk/ac/ebi/atlas/bioentity/geneset/GeneSetPropertyService.java
```diff
@@ -38,13 +38,13 @@ public GeneSetPropertyService(GoPoTrader goPoTermTrader,
     public Map<BioentityPropertyName, Set<String>> propertyValuesByType(String identifier) {
         identifier = identifier.toUpperCase();
         if (GeneSetUtil.matchesReactomeID(identifier)) {
```
wbazant: It's kind of surprising you didn't go for changing the signature of `propertyValuesByType` but it's not a problem.

gxa:
Passes build:
- http://bar:8085/browse/ATLAS-BASECIDEVTASKS11
- http://bar:8085/browse/ATLAS-GXACIDEVTASKS10

PR #24, opened 2017-10-24
Task 152021892

alfonsomunozpomer:
From fa07094 onwards the goal was to show species in orthologs and to put some order in the old `BioentityPropertyService`. The code there and surrounding classes was a bit messy.


alfonsomunozpomer:
Build passes tests:
- http://bar:8085/browse/ATLAS-BASECIDEVTASKS0-1
- http://bar:8085/browse/ATLAS-GXACIDEVTASKS14-1

base/src/main/java/uk/ac/ebi/atlas/search/SemanticQueryTerm.java
```diff
@@ -23,7 +23,7 @@ public static SemanticQueryTerm create(String value, String category) {
     public String asGxaIndexQueryLiteral() {
         return hasNoCategory()
                 ? String.format("property_value:\"%s\"", value())
-                : String.format("property_name:\"%s\" AND property_value:\"%s\"", category(), value().replace(":", "\\:").replace("[", "\\[").replace("]", "\\]"));
+                : String.format("property_name:\"%s\" AND property_value:\"%s\"", category(), value());
```
wbazant: I saw this on my phone and was concerned, I looked at it again and I'm not concerned about the change you did but I think we should escape the " that can potentially be inside value():
```
grep \" /ebi/ftp/pub/databases/microarray/data/atlas/bioentity_properties/ensembl/*.ensgene.tsv
``` 
returns some property values, and we don't tidy these values up on ingestion.

base/src/main/java/uk/ac/ebi/atlas/solr/query/SolrQueryService.java
```diff
@@ -74,4 +72,15 @@ public GeneQueryResponse fetchResponse(SemanticQuery geneQuery, Species species)
 
         return geneQueryResponse;
     }
+
+    public Set<String> fetchSpecies(SemanticQueryTerm geneQueryTerm) {
```
wbazant: This is confusing and doesn't look correct: why is the default property name "bioentity_identifier" ? If the category needs to be there, then why does the method take a SemanticQueryTerm that doesn't have to have a category?
Why are we not always delegating to `SemanticQueryTerm::asGxaIndexQueryLiteral` and adding this special case here?

base/src/main/java/uk/ac/ebi/atlas/solr/query/builders/BioentityIdentifierQueryBuilder.java
```diff
@@ -4,30 +4,42 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.ebi.atlas.search.SemanticQueryTerm;
-
-
-public class BioentityIdentifierQueryBuilder extends SolrQueryBuilder<BioentityIdentifierQueryBuilder>{
+import uk.ac.ebi.atlas.species.Species;
+
+/**
```
wbazant: Now that we have a better domain model (instead of stringly typed species and query term like they used to be) possibly it's time to ditch that builder. They're really Java's way of doing named or optional parameters (because constructors with, say, ten string parameters, are unmaintainable) but since this takes a Species and a SemanticQueryTerm there's no danger of switching them around (and we're handling allowing no species with a "Null object pattern" of unknown species).

base/src/main/java/uk/ac/ebi/atlas/utils/ReactomeClient.java
```diff
@@ -1,39 +1,108 @@
...
         try {
-            return Optional.of(StringUtils.trim(restTemplate.getForObject(url, String.class)));
+            return parseJsonResponse(restTemplate.postForObject(URL, postData, String.class), stableIds);
+        } catch (JsonSyntaxException e) {
+            LOGGER.error("Invalid JSON returned from Reactome API");
+            return ImmutableMap.of();
         } catch (RestClientException e) {
-            LOGGER.warn("Reactome ID \"{}\" could not be found", reactomeId);
-            return Optional.empty();
+            LOGGER.error("There was an error retrieving pathway names from Reactome");
```
wbazant: Use the error logs to tell the future you what the exception was, they'll be grateful.

base/src/test/java/uk/ac/ebi/atlas/utils/ReactomeClientTest.java
```diff
@@ -1,33 +1,129 @@
...
+        Map<String, String> result = subject.getPathwayNames(pathwayIds);
+
+        assertThat(result.keySet(), hasSize(3));
+        assertThat(result, hasEntry(is("R-FOO-111"), is("Bar signal")));
+        assertThat(result, hasEntry(is("R-FOO-222"), is("Bar receptor")));
+        assertThat(result, hasEntry(is("R-FOO-333"), is("Bar regulator")));
+    }
+
+    @Test
+    public void makesRequestsInBatchesOfQueryMaxSize() throws Exception {
```
wbazant: It's a minor point and this test is not incorrect, but you have done something TDD proponents rail against : you have used a property of an implementation (query 45 terms in batches of 20,20,5) instead of a requirement (query in batches of size no more than 20 because that is the max Reactome will take).

base/src/main/java/uk/ac/ebi/atlas/search/SemanticQueryTerm.java
alfonsomunozpomer: I couldn’t find any match in `analytics` with `identifier_search:*\"*` or `conditions_search:*\"*`, and none in `bioentities` with `property_value:*\"*`, but I think this will silence some of the nightly exceptions we get from bad syntax searches.

base/src/main/java/uk/ac/ebi/atlas/solr/query/SolrQueryService.java
alfonsomunozpomer: The problem is that if you look for, say, `ENSMUSG00000019082`, and use `asGxaIndexQueryLiteral` you will search in `property_value` and the first document that matches is:
```
{
        "bioentity_identifier":"WBGene00018901",
        "species":"Caenorhabditis_elegans",
        "property_value":"ENSMUSG00000019082",
        "property_weight":0,
        "property_name":"ortholog",
        "id":"8448a632-94b3-49e1-8bf3-dc95e6d8edd8",
        "_version_":1571896862507532314},
```

Then the species is resolved as *Caenorhabditis elegans*. This is specified in the the test `SolrQueryServiceTest::getKnownSpeciesWithoutCategory`.

The category isn’t mandatory, it may or may not be there, but if it isn’t what you want is to query `bioentity_identifier`. The one change I’m adding is to search first in `bioentity_identifier` and falling back to the regular `asGxaIndexQueryLiteral` if no results are returned.

PR #26, opened 2017-10-25
Task 152035624

wbazant:
Enable name for transcripts. While writing a test I noticed the code doesn't currently give right names to columns with one group. I added more logic to it, and the names should be all rightey now.


wbazant:
It's okay if this doesn't make it into the release.

base/src/test/java/uk/ac/ebi/atlas/profiles/stream/BaselineTranscriptProfileStreamFactoryTest.java
```diff
@@ -275,6 +261,19 @@ public void rightJson(){
...
+                                "                \"median\": 1.0,\n" +
+                                "                \"upper_quartile\": 1.0,\n" +
+                                "                \"max\": 1.0\n" +
+                                "            },\n" +
+                                "            \"values\": [{\n" +
+                                "                \"id\": \"assay_1\",\n" +
+                                "                \"assays\": [\"assay_1\"],\n" +
+                                "                \"value\": 1.0\n" +
+                                "            }]\n" +
                                 "        }],\n" +
```
alfonsomunozpomer: Just a side note: I’ve been thinking about this sort of tests lately, and I think that testing with JsonPath is definitely the way to go. It will make tests more robust and easier to modify when we e.g. add fields.

alfonsomunozpomer: Just a side note: I’ve been thinking about this sort of tests lately, and I think that testing with JsonPath is definitely the way to go. It will make tests more robust and easier to modify when we e.g. add fields.

Fix the test and it’s good.

PR #38, opened 2017-11-08
Public experiments ignore access key. Some changes.

wbazant:
Changes abound


base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentDAO.java
```diff
@@ -104,59 +95,36 @@ private void addExperimentSpeciesRows(ExperimentDTO experimentDTO) {
...
-        if (recordsCount == 0) {
-            throw new IllegalArgumentException("Experiment not found for accession " + experimentAccession);
-        }
+        int recordsCount = jdbcTemplate.update(UPDATE_EXPERIMENT, isPrivateAsString(isPrivate), experimentAccession);
+        checkState(recordsCount == 1 , "Experiment not found for accession " + experimentAccession );
 
     }
 
-    private String toString(boolean isPrivate) {
+    private String isPrivateAsString(boolean isPrivate) {
```
alfonsomunozpomer: I still wonder who thought that having the filed `private` be a `varchar(1)` instead of a `boolean` was a good idea. 😞

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentDAO.java
```diff
@@ -104,59 +95,36 @@ private void addExperimentSpeciesRows(ExperimentDTO experimentDTO) {
...
         int deletedRecordsCount = jdbcTemplate.update(DELETE_EXPERIMENT, experimentAccession);
-
-        if (deletedRecordsCount != 1) {
-            throw new IllegalArgumentException("Experiment not found for accession " + experimentAccession);
-        }
-    }
-
-    public ExperimentDTO findPublicExperiment(String experimentAccession) {
-        return findExperiment(experimentAccession, false);
+        checkState(deletedRecordsCount == 1 , "Experiment not found for accession " + experimentAccession);
```
alfonsomunozpomer: I would favour `checkArgument`: keeps the same exception semantics, `IllegalStateException` doesn’t make much sense here.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentDAO.java
```diff
@@ -104,59 +95,36 @@ private void addExperimentSpeciesRows(ExperimentDTO experimentDTO) {
...
-    public void updateExperiment(String experimentAccession, boolean isPrivate) {
-
-        int recordsCount = jdbcTemplate.update(UPDATE_EXPERIMENT, toString(isPrivate), experimentAccession);
+    public void setExperimentPrivacyStatus(String experimentAccession, boolean isPrivate) {
 
-        if (recordsCount == 0) {
-            throw new IllegalArgumentException("Experiment not found for accession " + experimentAccession);
-        }
+        int recordsCount = jdbcTemplate.update(UPDATE_EXPERIMENT, isPrivateAsString(isPrivate), experimentAccession);
+        checkState(recordsCount == 1 , "Experiment not found for accession " + experimentAccession );
```
alfonsomunozpomer: See https://github.com/gxa/atlas/pull/38/commits/6f932fc501807c4eae09db3e7c4b4df50727bc21#r151509665

base/src/main/java/uk/ac/ebi/atlas/trader/ExperimentTrader.java
```diff
@@ -16,7 +16,9 @@ public ExperimentTrader(ExperimentDAO experimentDAO) {
         this.experimentDAO = experimentDAO;
     }
 
-    public abstract Experiment getPublicExperiment(String experimentAccession);
+    public Experiment getPublicExperiment(String experimentAccession) {
+        return getExperiment(experimentAccession, "");
+    }
```
alfonsomunozpomer: I don’t understand this change...

base/src/test/java/uk/ac/ebi/atlas/experimentimport/ExperimentDAOIT.java
```diff
@@ -32,24 +35,29 @@
     @Inject
     private ExperimentDAO subject;
 
-    public void createSecret111() throws Exception {
-        ExperimentDTO mtab = ExperimentDTO.createNew(SECRET_111, TYPE_MICROARRAY, "cow", Sets.newHashSet("1"), "diff", false);
-        subject.addExperiment(mtab, Optional.<String>absent());
+    public UUID createSecret111(boolean isPrivate) {
+        ExperimentDTO mtab = ExperimentDTO.createNew(SECRET_111, TYPE_MICROARRAY, "cow", Sets.newHashSet("1"), "diff", isPrivate);
+        return subject.addExperiment(mtab, Optional.<String>absent());
```
alfonsomunozpomer: You don’t need to specify `<String>` here, Java 8 will infer it for you.

base/src/test/java/uk/ac/ebi/atlas/experimentimport/ExperimentDAOIT.java
```diff
@@ -58,53 +66,89 @@ public void testFindExperimentByType() throws Exception {
...
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void cannotCreateExperimentTwice2(){
+        createSecret111(false);
+        createSecret111(true);
+    }
+
+    @Test
+    public void forPublicExperimentsIdIsIgnored()  {
```
alfonsomunozpomer: `Id` → `accessKey`. Same below in `forPrivateExperimentsIdIsRequired`.

gxa/src/main/java/uk/ac/ebi/atlas/trader/ExpressionAtlasExperimentTrader.java
```diff
@@ -52,33 +47,27 @@ public ExpressionAtlasExperimentTrader(ExperimentDAO experimentDAO,
...
-            return getExperimentFromCache(experimentAccession, experimentType);
+            return getExperimentFromCache(experimentAccession, experimentTypes.get(Pair.of(experimentAccession, accessKey)));
         } catch (ExecutionException e) {
-            LOGGER.error(e.getMessage());
-            throw new ResourceNotFoundException("Failed to load <em>" + experimentAccession + "</em>.");
-        }
-        catch (UncheckedExecutionException | NullPointerException e) {
-            LOGGER.error(e.getMessage());
-            throw new ResourceNotFoundException("Experiment <em>" + experimentAccession + "</em> not found.");
+            throw new RuntimeException(e);
```
alfonsomunozpomer: `ResourceNotFoundException` returns a 404 in `...ExceptionHandlingController`, whereas this makes it return a generic 400.

wbazant: This is a really good point. I've changed the DAO to consistently throw the ResourceNotFoundException there, and the LoadingCache currently catches exceptions and rethrows them as UncheckedExecutionException - I'll try to do something about it

base/src/main/java/uk/ac/ebi/atlas/trader/ExperimentTrader.java
wbazant: This is a convenience method. We can implement it right up in ExperimentTrader - despite not knowing what getExperiment will do exactly, we expect it to give us a public experiment with "" for an accessKey, so we do that instead of marking it as abstract and getting the children to all do it in their own slightly different way.

wbazant:
@alfonsomunozpomer I think it's good now - before I tried to avoid treating all exceptions (like: missing file) as "no experiment was found at all" but you made a very good point about using ResourceNotFoundException when there is really no experiment. I think the current code achieves both of these.

PR #43, opened 2017-11-27
Task 148227051 (Update the /das/s4/features endpoint)

haideriqbal:
https://www.pivotaltracker.com/story/show/148227051

- Added new controller according to problem specification.
- Added new json result parser class to parse the results and map them to Differential Analytic objects.
- Created test cases for both of the classes mentioned above
- New controller is not in its final state as we need to compare the output of both new and old controller to make sure they are same.

### DIFF
--- a/gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
+++ b/gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
@@ -1,0 +1,126 @@
+package uk.ac.ebi.atlas.controllers.rest;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.SetMultimap;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import org.springframework.context.annotation.Scope;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.servlet.ModelAndView;
+import uk.ac.ebi.atlas.controllers.JsonExceptionHandlingController;
+import uk.ac.ebi.atlas.model.AssayGroup;
+import uk.ac.ebi.atlas.model.experiment.Experiment;
+import uk.ac.ebi.atlas.model.experiment.baseline.BioentityPropertyName;
+import uk.ac.ebi.atlas.model.experiment.baseline.Factor;
+import uk.ac.ebi.atlas.model.experiment.baseline.impl.FactorSet;
+import uk.ac.ebi.atlas.model.experiment.differential.Contrast;
+import uk.ac.ebi.atlas.model.experiment.differential.DifferentialExpression;
+import uk.ac.ebi.atlas.model.experiment.differential.microarray.MicroarrayExpression;
+import uk.ac.ebi.atlas.search.SemanticQuery;
+import uk.ac.ebi.atlas.search.SemanticQueryTerm;
+import uk.ac.ebi.atlas.search.analyticsindex.differential.DifferentialAnalyticsSearchDao;
+import uk.ac.ebi.atlas.search.analyticsindex.differential.DifferentialAnalyticsSearchService;
+import uk.ac.ebi.atlas.search.diffanalytics.DiffAnalytics;
+import uk.ac.ebi.atlas.search.diffanalytics.DiffAnalyticsQueryBuilder;
+import uk.ac.ebi.atlas.search.diffanalytics.DiffAnalyticsSearchService;
+import uk.ac.ebi.atlas.species.SpeciesFactory;
+import uk.ac.ebi.atlas.trader.ContrastTrader;
+import uk.ac.ebi.atlas.trader.ExperimentTrader;
+
+import javax.inject.Inject;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+@Controller
+@Scope("request")
+public class DASFeaturesControllerNew {
+
+
+    private ExperimentTrader experimentTrader;
+    private DifferentialAnalyticsSearchService differentialAnalyticsSearchService;
+    private DifferentialJsonResultsParser differentialJsonResultsParser;
+
+
+    @Inject
+    public DASFeaturesControllerNew(ExperimentTrader experimentTrader,
+                                    DifferentialAnalyticsSearchService differentialAnalyticsSearchService,
+                                    DifferentialJsonResultsParser differentialJsonResultsParser) {
+
+        this.experimentTrader = experimentTrader;
+        this.differentialAnalyticsSearchService = differentialAnalyticsSearchService;
+        this.differentialJsonResultsParser = differentialJsonResultsParser;
+    }
+
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
+    public ModelAndView InternalServerHandleException(Exception e) {
+        ModelAndView mav = new ModelAndView("error-page");
+        mav.addObject("exceptionMessage", e.getMessage());
+        return mav;
+    }
+
+    @RequestMapping(value = "/das/s4/features/test")
+    public String getDifferentialJsonResults(@RequestParam(value = "segment", required = false, defaultValue = "")
+                                                     String geneId,
+                                             @RequestParam(value = "conditionQuery", required = false, defaultValue = "")
+                                                     SemanticQuery conditionQuery,
+                                             Model model) {
+
+        try {
+
+            List<DiffAnalytics> diffAnalyticsList;
+            SetMultimap<String, String> factorValuesByType = HashMultimap.create();
+
+            JsonObject result = differentialAnalyticsSearchService.fetchResults(SemanticQuery.create(geneId), conditionQuery);
+
+            diffAnalyticsList = differentialJsonResultsParser.parseDifferentialResults(result);
+
+            String geneName = diffAnalyticsList.isEmpty() ? geneId : diffAnalyticsList.get(0).getBioentityName();
+
+            for (DiffAnalytics dbe : diffAnalyticsList) {
+                AssayGroup testAssayGroup = dbe.getContrastTestAssayGroup();
+                Experiment experiment = experimentTrader.getPublicExperiment(dbe.getExperimentAccession());
+
+                FactorSet factorsForAssayGroup = FactorSet.create(experiment.getExperimentDesign().getFactorValues(testAssayGroup.getFirstAssayAccession()));
+                for (Factor factor : factorsForAssayGroup) {
+                    factorValuesByType.put(factor.getType(), factor.getValue());
+                }
+            }
+
+            model.addAttribute("geneId", geneId);
+            model.addAttribute("geneName", geneName);
+            model.addAttribute("factorValues_ORGANISM_PART", formatFactorValuesLabel(factorValuesByType.get("ORGANISM_PART")));
+            model.addAttribute("factorValues_DISEASE", formatFactorValuesLabel(factorValuesByType.get("DISEASE")));
+            model.addAttribute("factorValues_CELL_TYPE", formatFactorValuesLabel(factorValuesByType.get("CELL_TYPE")));
+            model.addAttribute("factorValues_CELL_LINE", formatFactorValuesLabel(factorValuesByType.get("CELL_LINE")));
+            model.addAttribute("factorValues_COMPOUND", formatFactorValuesLabel(factorValuesByType.get("COMPOUND")));
+            model.addAttribute("factorValues_DEVELOPMENTAL_STAGE", formatFactorValuesLabel(factorValuesByType.get("DEVELOPMENTAL_STAGE")));
+            model.addAttribute("factorValues_INFECT", formatFactorValuesLabel(factorValuesByType.get("INFECT")));
+            model.addAttribute("factorValues_PHENOTYPE", formatFactorValuesLabel(factorValuesByType.get("PHENOTYPE")));
+
+            return "das-features";
+        }
+        catch (Exception e)
+        {
+            throw e;
+        }
+    }
+
+    private String formatFactorValuesLabel(Set<String> factorValues) {
+        if (factorValues.isEmpty()) {
+            return "Not studied or no differential expression found for this gene";
+        }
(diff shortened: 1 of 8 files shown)


gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+import uk.ac.ebi.atlas.trader.ContrastTrader;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.Optional;
+
+@Named
+public class DifferentialJsonResultsParser extends JsonExceptionHandlingController {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DifferentialResultsReader.class);
```
alfonsomunozpomer: The logger isn’t used at all. It might be good to log some messages? However observe that the reporting class will be `DifferentialResultsReader`, which is incorrect.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private ContrastTrader contrastTrader;
+
+    @Inject
+    public DifferentialJsonResultsParser(ContrastTrader contrastTrader)
+    {
```
alfonsomunozpomer: We don’t have a style guide, but the project follows loosely the [Java K&R style](https://en.wikipedia.org/wiki/Indentation_style#Variant:_Java). It’s not super important but try to follow the style of the project.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+        try
+        {
+            ImmutableList.Builder<DiffAnalytics> diffAnalyticsListBuilder = ImmutableList.builder();
+            JsonArray results = obj.getAsJsonArray("results");
+
+            for(JsonElement e : results) {
+
+                DiffAnalytics diffAnalytics = buildDiffAnalyticObject(e);
+
+                if(diffAnalytics != null)
```
alfonsomunozpomer: If you’re returning `null`, and it’s not an error, `Optional` is usually preferable. It expresses clearly that an absent value is ok, and that you should act accordingly.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+
+            return diffAnalyticsListBuilder.build();
+        }
+        catch (RuntimeException e)
+        {
+            throw new RuntimeException("Input Json is missing a results field OR results field is not an Array",e);
+        }
+    }
+
+    private DiffAnalytics buildDiffAnalyticObject(JsonElement e) {
```
alfonsomunozpomer: Typo: it’s *analytics*.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+                    buildDifferentialExpression(
+                            resultObj.get("p_value").getAsDouble(),
+                            resultObj.get("fold_change").getAsDouble(),
+                            Optional.ofNullable(resultObj.get("t_statistic")).map(JsonElement::getAsDouble).orElse(null));
+
+            String species = resultObj.get("species").getAsString();
+            species = species.substring(0,1).toUpperCase() + species.substring(1);
+
+
+            return new DiffAnalytics(resultObj.get("bioentity_identifier").getAsString(), Optional.ofNullable(resultObj.get("keyword_symbol")).map(JsonElement::getAsString).orElse(resultObj.get("bioentity_identifier").getAsString()), resultObj.get("experiment_accession").getAsString(), expression, species, contrast);
```
alfonsomunozpomer: Notice that you’re parsing more than once the same values in the same line (e.g. `bioentity_identifier`) or within this method (e.g. `experiment_accession`).

I would first deserialize the JSON object into its primitives and then build the `DiffAnalytics` object. Also, that would solve kludge below for building the `DifferentialExpression`: deserializing `t_statistic` as an `Optinal` and use one constructor or the other.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -0,0 +1,93 @@
...
+
+                if(diffAnalytics != null)
+                {
+                    diffAnalyticsListBuilder.add(diffAnalytics);
+                }
+            }
+
+            return diffAnalyticsListBuilder.build();
+        }
+        catch (RuntimeException e)
```
alfonsomunozpomer: A `RuntimeException` isn’t very helpful. Can you catch different types and display a specific message instead of "This or that"? This will also make your testing more succinct by using the syntax `@Test(expected=MyException.class)`  without needing to care about the message.

gxa/src/test/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParserTest.java
```diff
@@ -0,0 +1,186 @@
...
+            element.addProperty("experiment_accession", "e"+ i);
+            element.addProperty("contrast_id", "c"+ i);
+            element.addProperty("bioentity_identifier", "id"+ i);
+            element.addProperty("species", "org"+ i);
+            element.addProperty("p_value", 1.11);
+            element.addProperty("fold_change", -1.12);
+            element.addProperty("t_statistic", 0.0);
+            element.addProperty("keyword_symbol", i);
+            results.add(element);
+        }
```
alfonsomunozpomer: Refactor this loop into a utilitiy method that returns synthetic test data of any given size instead.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -0,0 +1,126 @@
...
+            model.addAttribute("factorValues_CELL_TYPE", formatFactorValuesLabel(factorValuesByType.get("CELL_TYPE")));
+            model.addAttribute("factorValues_CELL_LINE", formatFactorValuesLabel(factorValuesByType.get("CELL_LINE")));
+            model.addAttribute("factorValues_COMPOUND", formatFactorValuesLabel(factorValuesByType.get("COMPOUND")));
+            model.addAttribute("factorValues_DEVELOPMENTAL_STAGE", formatFactorValuesLabel(factorValuesByType.get("DEVELOPMENTAL_STAGE")));
+            model.addAttribute("factorValues_INFECT", formatFactorValuesLabel(factorValuesByType.get("INFECT")));
+            model.addAttribute("factorValues_PHENOTYPE", formatFactorValuesLabel(factorValuesByType.get("PHENOTYPE")));
+
+            return "das-features";
+        }
+        catch (Exception e)
```
wbazant: What's that for?

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -0,0 +1,126 @@
...
+
+                FactorSet factorsForAssayGroup = FactorSet.create(experiment.getExperimentDesign().getFactorValues(testAssayGroup.getFirstAssayAccession()));
+                for (Factor factor : factorsForAssayGroup) {
+                    factorValuesByType.put(factor.getType(), factor.getValue());
+                }
+            }
+
+            model.addAttribute("geneId", geneId);
+            model.addAttribute("geneName", geneName);
+            model.addAttribute("factorValues_ORGANISM_PART", formatFactorValuesLabel(factorValuesByType.get("ORGANISM_PART")));
```
wbazant: Consider not repeating yourself:
```
for(String factorValue: ImmutableList.of("ORGANISM_PART", "DISEASE"... ) {
model.addAttribute("factorValues_"+factorValue, formatFactorValuesLabel(factorValuesByType.get(factorValue));
}
```

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -0,0 +1,126 @@
...
+                                    DifferentialAnalyticsSearchService differentialAnalyticsSearchService,
+                                    DifferentialJsonResultsParser differentialJsonResultsParser) {
+
+        this.experimentTrader = experimentTrader;
+        this.differentialAnalyticsSearchService = differentialAnalyticsSearchService;
+        this.differentialJsonResultsParser = differentialJsonResultsParser;
+    }
+
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
+    public ModelAndView InternalServerHandleException(Exception e) {
```
wbazant: I like that this exists. Consider, though, that the endpoint is in XML, and it's used programmatically - probably the client will parse it to get individual fields - maybe it's better to show an error message in the same kind of format?
Also maybe it's not the best to show external users our internal messages - how about logging the error message in our logs, and saying something like "Unexpected error occurred, email us if this persists"?

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -0,0 +1,126 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+@Controller
+@Scope("request")
+public class DASFeaturesControllerNew {
+
+
+    private ExperimentTrader experimentTrader;
```
wbazant: `private final` please (or at least, next time around).

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
haideriqbal: Will keep that in mind for future.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
haideriqbal: Used Optional in this case.

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -65,7 +49,8 @@ public DASFeaturesControllerNew(ExperimentTrader experimentTrader,
     @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
     public ModelAndView InternalServerHandleException(Exception e) {
         ModelAndView mav = new ModelAndView("error-page");
-        mav.addObject("exceptionMessage", e.getMessage());
+        mav.addObject("exceptionMessage", "Unexpected error occurred, email us if this presists.");
+        LOGGER.error("Error getting Differential json results.",e.getMessage());
```
alfonsomunozpomer: Include the full `e` in the error. You saw how masking the stack trace gave us trouble in debugging the Reactome HTTPS error. #LearningTheHardWay

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DASFeaturesControllerNew.java
```diff
@@ -99,21 +82,12 @@ public String getDifferentialJsonResults(@RequestParam(value = "segment", requir
...
-            model.addAttribute("factorValues_DEVELOPMENTAL_STAGE", formatFactorValuesLabel(factorValuesByType.get("DEVELOPMENTAL_STAGE")));
-            model.addAttribute("factorValues_INFECT", formatFactorValuesLabel(factorValuesByType.get("INFECT")));
-            model.addAttribute("factorValues_PHENOTYPE", formatFactorValuesLabel(factorValuesByType.get("PHENOTYPE")));
 
-            return "das-features";
-        }
-        catch (Exception e)
-        {
-            throw e;
+        for (String factorValue: ImmutableList.of("ORGANISM_PART","DISEASE","CELL_TYPE","CELL_LINE","COMPOUND","DEVELOPMENTAL_STAGE","INFECT","PHENOTYPE")) {
```
alfonsomunozpomer: +1

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -31,63 +32,69 @@ public DifferentialJsonResultsParser(ContrastTrader contrastTrader)
...
                     diffAnalyticsListBuilder.add(diffAnalytics);
                 }
             }
 
             return diffAnalyticsListBuilder.build();
         }
-        catch (RuntimeException e)
-        {
-            throw new RuntimeException("Input Json is missing a results field OR results field is not an Array",e);
+        catch (IllegalFormatException e) {
```
alfonsomunozpomer: I don’t think this is the exception you want. From https://docs.oracle.com/javase/8/docs/api/java/util/IllegalFormatException.html:
> Unchecked exception thrown when a format string...

A format string is what’s used by `String::format`. How about a [JsonParseException](https://google.github.io/gson/apidocs/com/google/gson/JsonParseException.html)?

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/DifferentialJsonResultsParser.java
```diff
@@ -31,63 +32,69 @@ public DifferentialJsonResultsParser(ContrastTrader contrastTrader)
...
             String species = resultObj.get("species").getAsString();
             species = species.substring(0,1).toUpperCase() + species.substring(1);
 
-
-            return new DiffAnalytics(resultObj.get("bioentity_identifier").getAsString(), Optional.ofNullable(resultObj.get("keyword_symbol")).map(JsonElement::getAsString).orElse(resultObj.get("bioentity_identifier").getAsString()), resultObj.get("experiment_accession").getAsString(), expression, species, contrast);
+            return new DiffAnalytics(bioentity_identifier, keyword_symbol, experiment_accession, expression, species, contrast);
         }
-        catch(RuntimeException exp)
-        {
+        catch(NullPointerException exp) {
```
alfonsomunozpomer: Which method is likely to throw a NPE? This is a down-to-earth, honest question.

PR #47, opened 2017-12-01
Task 152395202

alfonsomunozpomer:
* Make `DataFileHub` not be concerned about species properties
* Add Matrix Market parser
* Add Matrix Market expression files to `DataFileHub`
* Make a `SingleCellAnalytics` class and factory that is plugged to the parser
* Split infrastructure for GXA and SCXA (both in the DB and in the data directory)
* Added all the plumbing to have a good separation between experiment admin classes for GXA and SCXA

The new directory structure needed to make this work is (already fixed in `lime`):
```
.
└── integration-test-data
    ├── bioentity_properties -> <atlas-dir>/base/src/test/resources/data-files/bioentity_properties/
    ├── gxa
    │   ├── admin
    │   ├── expdesign -> <atlas-dir>/base/src/test/resources/data-files/experiments_gxa/expdesign/
    │   ├── magetab -> <atlas-dir>/base/src/test/resources/data-files/experiments_gxa/magetab/
    │   └── serialized_expression
    ├── ontology -> <atlas-dir>/base/src/test/resources/data-files/ontology/
    ├── rdf
    ├── scxa
    │   ├── admin
    │   ├── expdesign -> <atlas-dir>/base/src/test/resources/data-files/experiments_scxa/expdesign/
    │   └── magetab -> <atlas-dir>/base/src/test/resources/data-files/experiments_scxa/magetab/
    ├── solr -> <atlas-dir>/base/src/test/resources/solr/
    └── species -> <atlas-dir>/base/src/test/resources/data-files/species/
```

Passes all tests, old and new:
* http://bar:8085/browse/ATLAS-BASECIDEVTASKS1-13
* http://bar:8085/browse/ATLAS-GXACIDEVTASKS0-9
* http://bar:8085/browse/ATLAS-SCXACIDEVTASKS2-10


alfonsomunozpomer:
• `SingleCellAnalytics::toCsvLine` ☑

• The *GoF* hashtag was a bit of a joke. Mostly because many times the pattern is thought to be just a class with a `create` method of sorts, whereas the formal original definition is to create concrete classes of a common interface (in this case with three interfaces). I’m not a purist and I think that depending the term is often abused, but that’s alright if that’s understood. In terms of design, the factory isn’t producing singletons, it’s creating an `ExperimentCrud` with each request to `/admin/experiments`. This was before I decided to rewrite `UrlTsvReaderBuilder` which had mutable state, so maybe now the criticism is fair because we can have two `ExperimentCrud` singletons (going to the previous design and injecting them in the controllers). Still, I think the factory makes it clear what collaborators are common and which ones modify the behaviour of the concrete class. It also helps when testing `ExperimentCrud` at least as it is.

• `SingleCellAnalyticsStream` has the following job: take three files (Matrix Market file with column and row labels) and streams the matrix entries decorated with the labels as. Maybe a better design is to have a class that just does that and a second one that maps the stream to `SingleCellAnalytics`?

• `MatrixMarketFile` ☑

• I struggled with the triple a bit... I’m open to suggestions. The crux of the matter is that those three files should go together. They don’t make sense on their own, really. Maybe if we `SingleCellAnalyticsStream` is decoupled from the three files then these three files can be their own thing, too.

PR #48, opened 2017-12-01
Task 152400453 2

alfonsomunozpomer:
After the first proof-of-concept, the meaty part starts at https://github.com/gxa/atlas/commit/8d951408d6a6b1df7432782853be6e05d61c4e2c. You can verify results using the benchmark class. The fact that the baseline profiles built now are incorrect makes it hard to compare against the current methods.


base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+            this.name = name;
+        }
+
+        @Override
+        public String toString() {
+            return this.name;
+        }
+    }
+
+    private static final String SIMPLE_QUERY_TEMPLATE = "{0}:\"{1}\"";
```
wbazant: I like this class!

As far as I see you're only using these constants in one place each, you could controversially put them there and save everyone some scrolling.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final String TERMS_QUERY_TEMPLATE = "('{'!terms f={0}'}'{1})";
+    private static final String RANGE_QUERY_TEMPLATE = "{0}:[{1} TO *]";
+
+    // Solr defaults
+    private HashSet<String> filterQueries = Sets.newHashSet();
```
wbazant: This could be an ImmutableSet.Builder if you fancy.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+
+        @Override
+        public String toString() {
+            return this.name;
+        }
+    }
+
+    private static final String SIMPLE_QUERY_TEMPLATE = "{0}:\"{1}\"";
+    private static final String TERMS_QUERY_TEMPLATE = "('{'!terms f={0}'}'{1})";
+    private static final String RANGE_QUERY_TEMPLATE = "{0}:[{1} TO *]";
```
wbazant: You're using the word "query" to mean the result of this class - an analytics query - and also individual bits here that the query later consists of. Rename these as "clauses", like in SQL?

wbazant: Or: "filters"?

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsDao.java
```diff
@@ -0,0 +1,125 @@
...
+                                                                       double expressionLevelThreshold) {
+        AnalyticsQueryBuilder queryBuilder = new AnalyticsQueryBuilder();
+        queryBuilder
+                .filterExperimentAccession(experimentAccession)
+                .filterAssayGroupIds(sliceAssayGroupIds)
+                .queryExpressionLevel(expressionLevelThreshold)
+                .queryNotAssayGroupIds(selectedAssayGroupIds)
+                .setJsonFacet(maxExpressionOverBioentityIdentifiersJsonFacet);
+
+        if (geneIds.size() < TERMS_QUERY_SIZE_THRESHOLD) {
```
wbazant: This looks dangerous. If there are too many genes, do we filter them out from results afterwards? Maybe a comment should say where that place is?

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsSearchService.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static uk.ac.ebi.atlas.solr.analytics.fullanalytics.FullAnalyticsSearchService.SpecificGeneProfile.COMPARATOR_BY_SPECIFICITY_AND_NAIVE_DIFF_EXPRESSION;
+
+@Named
+public class FullAnalyticsSearchService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FullAnalyticsSearchService.class);
```
wbazant: Unused

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsSearchService.java
```diff
@@ -0,0 +1,182 @@
...
+        return geneProfilesInSelectedAssayGroupsWithSelectionData.stream()
+                .map(profile ->
+                        SpecificGeneProfile.merge(
+                                profile,
+                                geneProfilesInSelectedAssayGroupsWithSliceDataByGeneId.get(profile.geneId())))
+                .collect(Collectors.toSet());
+    }
+
+    @AutoValue
+    static abstract class SpecificGeneProfile {
```
wbazant: I was very confused by this class, and I didn't understand what it represents, and then I understood that it represents some partial results, and then also the end result you want just before sorting.

Also this class is coupled to AnalyticsDao, and the payload received. I recommend you:
- represent the output of AnalyticsDao by Map<String, Pair<Integer,Double>> (gene id, count, statistic value [max /avg expression])  or something convenient 
- move the logic of creating that to AnalyticsDao
- create `SpecificGeneProfile` from the map entries perhaps, by some extra constructors? 
- remove the merge method from SpecificGeneProfile, and replace with putting together the two maps in `specificExpressionInSliceAcrossSelectedAssayGroups`

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsSearchService.java
```diff
@@ -0,0 +1,182 @@
...
+        static final Comparator<SpecificGeneProfile> COMPARATOR_BY_SPECIFICITY_AND_NAIVE_DIFF_EXPRESSION = (o1, o2) ->
+            ComparisonChain.start()
+                    .compare(
+                            o1.countsInSelection() + o1.countsOutsideSelection(),
```
wbazant: After AnalyticsDao is refactored to provide an interface we can work with, we can create beautiful, beautiful tests for this comparator - it's now quite hard for me to assert the correctness of this method, although it seems to be doing a couple of things I recall from elsewhere.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/query/AnalyticsQueryClient.java
```diff
@@ -36,41 +37,41 @@
...
+            LOGGER.debug("fetchResults {} took {} seconds", q, stopwatch.elapsed(TimeUnit.MILLISECONDS) / 1000D);
+            if (responseNonEmpty(result)) {
                 break;
             }
         }
         return result;
     }
 
-    boolean responseNonEmpty(String jsonFromSolr){
+    protected boolean responseNonEmpty(String jsonFromSolr) {
```
wbazant: I see no reason for you to have changed this and it's only used in fetchResults - make private?

base/src/test/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilderTest.java
```diff
@@ -0,0 +1,27 @@
+package uk.ac.ebi.atlas.solr.analytics.fullanalytics;
+
+import org.apache.solr.client.solrj.SolrQuery;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.isA;
+import static org.junit.Assert.assertThat;
+
+public class AnalyticsQueryBuilderTest {
```
wbazant: A quick test that would nevertheless be verifying something and not giving false positives would be when set a field value to be a magic value - say a cutoff is 133.7 -  and assert it appears somewhere in the result as a substring.

base/src/test/java/uk/ac/ebi/atlas/utils/StringUtilTest.java
```diff
@@ -1,23 +1,36 @@
...
 
     @Test
-    public void quoteIfMoreThanOneWord() throws Exception {
+    public void quoteIfMoreThanOneWord() {
         assertThat(StringUtil.quoteIfMoreThanOneWord("foo bar"), is("\"foo bar\""));
         assertThat(StringUtil.quoteIfMoreThanOneWord("foobar"), is("foobar"));
     }
 
+    @Test
+    public void escapeDoubleQuotes() {
```
wbazant: You're testing a library?

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilder.java
alfonsomunozpomer: Clause seems better, filter can be mis-leading, being this a Solr-related class.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsDao.java
alfonsomunozpomer: So, if the number of genese is very big there’s a performance hit (I don’t know if it’s the network or Solr figuring out the query), so getting all gene IDs is better. In a way, `geneIds` is helping Solr to filter results so we don’t need to. I’ll add a comment to document this.

base/src/test/java/uk/ac/ebi/atlas/utils/StringUtilTest.java
alfonsomunozpomer: It’s *our* library! :)

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/FullAnalyticsSearchService.java
alfonsomunozpomer: Very fair point, the coupling is a concern and I went back and forth with several ideas. At one point it was part of the DAO, then it was an independent thing, and it ended as it is. I agree it’s not pretty.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/query/AnalyticsQueryClient.java
alfonsomunozpomer: At some point I was reusing those methods in `AnalyticsClient`. At some point I think we’ll need to integrate the two classes. Thanks for spotting it.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/fullanalytics/AnalyticsQueryBuilder.java
alfonsomunozpomer: After reading Solr’s reference guide [the most widespread term in Solr parlance seems to be “term”](https://lucene.apache.org/solr/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-SpecifyingTermsfortheStandardQueryParser). I’ll go with that.

PR #50, opened 2017-12-06
Task 150050573

elisabetb:
Integrating bioentity gene Information into sc experiment page


elisabetb:
I have run build manually and the test are passing:
http://bar:8085/browse/ATLAS-GXACIDEVTASKS-72

wbazant:
I have looked at the code, doesn't seem harmful -mostly because you've duplicated things or likely to cause trouble so merging. Do see if it shows up in www-test after the build.

PR #51, opened 2017-12-08
Task 148227051 1

wbazant:
I have:
- reverted the revert
- cherry-picked the head commit of the branch task148227051
I think this is all the changes? @haideriqbal do have a look if it's the pull request you intended to make.

I was confused by one commit before task148227051 - thought it's a new commit you made, only later looked at the date and it's the one you've addressed before.

Btw we have a naming convention for pull requests - task_<task id> or task_<task id>_<number> for multiple pull requests per task.

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/search/diffanalytics/DiffAnalytics.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/search/diffanalytics/DiffAnalytics.java
@@ -36,10 +36,6 @@ public class DiffAnalytics {
         return bioentityId;
     }
 
-    public String getContrastDisplayName() {
-        return contrast.getDisplayName();
-    }
-
     public AssayGroup getContrastTestAssayGroup() {
         return contrast.getTestAssayGroup();
     }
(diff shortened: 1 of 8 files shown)


wbazant:
Can you review my changes?
- Use camel case for variable names
- One optional a bit different

I have also more controversially removed the two try catch blocks. Have you seen during development the case where the NullPointerException is thrown? It's really bad and we need to fix it if so.

haideriqbal:
@wbazant yes there is a case when method throws a NullPointerException. If we are trying to access the field from json which is not there, then the method throws NullPointerException and to skip the faulty element I've catch the exception and logged the error.

wbazant:
@haideriqbal I think all the records should come from Solr in exactly the same format. So I think there's never going to be a case where the application is working properly, and you get both some valid records and invalid ones - if there are NPEs I except all records to be wrong, so I think it's right to treat it as a fatal error (crash, show the error page, let the developers know about the error) instead of a nonfatal one (log the error, and try to show partial result). 

That's why I'm asking - have you seen during development a case where the records did come back in a different format? I understand the t-statistic field only being there for microarray experiments - and your code handles this - but for example, have you ever seen a case where experiment accession is sometimes missing in the record and the code needs to skip it?

PR #53, opened 2018-02-15
Add support for Plant Reactome Ids

haideriqbal:
Requirements as stated in this story are implemented:
https://www.pivotaltracker.com/story/show/152441439

Modify Reactome client to handle plant pathwayids
Modify GenesetPropertyService to distinguish plant and non plant species


gxa:
Please provide links to green Bamboo builds!

haideriqbal:
Links to green builds:

http://bar:8085/browse/ATLAS-BASECIDEVTASKS2-1
http://bar:8085/browse/ATLAS-GXACIDEVTASKS2

alfonsomunozpomer:
Ok, that’s a promising start! I think it‘s the minimal set of code changes and we won’t need to change anything in the annotation process. However, I have a few questions:

1. Have you tested that the feature works? What are the expected results?
2. What are the scenarios where this feature is going to be shown?

A technicality, but also important: code coverage in `ReactomeClient` has decreased from 100% to 86%.

I think the new test added in `SpeciesInferrerIT` should be a separate one, at least for the sake of clarity. If I look at the commit I don’t know what exactly is the importance of that particular assertion. Also, be granular about your commits. Making a big commit with a message that sums up the feature isn’t useful, the feature is already identified by the branch number.

haideriqbal:
Agreed with plant id test in SpeciesInferrerIT to be separate. I would do it.
Regarding commit message I will be more granular.
In aspects of scenarios of where this feature is useful. I will look into code and will see where this feature is being used.

PR #67, opened 2018-03-14
Task 150050573 - Bioentity information endpoint

monicajianu:
- Added new /json/bioentity_information/<gene_id> endpoint which returns bioentity properties given an identifier
- Tested new controller
- Added sorting by relevance to property values
- Added new loading gif asset to be used by React component


base/src/main/java/uk/ac/ebi/atlas/bioentity/properties/BioEntityCardModelFactory.java
```diff
@@ -158,6 +148,10 @@ private String getFirstValueOfProperty(BioentityPropertyName propertyType,
                                 )
                 ).collect(Collectors.toList());
 
+        // Sort by relevance in descending order
+        links.sort((PropertyLink p1, PropertyLink p2) -> p2.getRelevance() - p1.getRelevance());
+
+        return links;
```
alfonsomunozpomer: Java streams have a [`.sorted`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#sorted-java.util.Comparator-) method that you can chain, so this could be:
```
return bioEntityPropertyService.mapToLinkText(propertyName, propertyValues).entrySet().stream()
    ...
    .sorted((p1, p2) -> p2.getRelevance() - p1.getRelevance())
    .collect(toList())
```

The comparator can further be simplified to `.sorted(comparing(PropertyLink::getRelevance))` but check docs first.

sc/src/main/java/uk/ac/ebi/atlas/search/JsonBioentityInformationController.java
```diff
@@ -51,15 +43,8 @@ public String getBioentityInformation(@PathVariable String geneId) {
                 bioEntityPropertyDao.fetchPropertyValuesForGeneId(geneId, BioentityPropertyName.SYMBOL).stream()
                         .collect(Collectors.joining("/"));
 
-
         Map<String, Object> model = bioEntityCardModelFactory.modelAttributes(geneId, species, BioEntityCardProperties.bioentityPropertyNames, geneName, propertyValues);
 
         return model.get("bioentityProperties").toString();
```
alfonsomunozpomer: So far we’ll leave it there. It’s sort of ugly but it’ll do. Ideally we can refactor the `bioentityProperties` and explicitly JSON-serialize it instead of having a `toString` method which is semantically confusing. This will become its own thing when we deprecate the current model and both EA and SCEA use this endpoint and the new component.

monicajianu:
@alfonsomunozpomer The Bamboo builds for the Single Cell module have been failing [for a while now](http://bar.ebi.ac.uk:8085/browse/ATLAS-SCXACIDEVTASKS )... I think some of the other tests have issues, I could have a look at them.

PR #70, opened 2018-03-22
Task 1551186043 - Idf parser retrieves Array Express display name

monicajianu:
* Idf parser retrieves Array Express display name, if that does not exist, it uses the investigation title
* Fixed existing tests
* Added aditional test case

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentimport/condensedSdrf/IdfParser.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentimport/condensedSdrf/IdfParser.java
@@ -12,6 +12,7 @@ import java.util.Arrays;
 public class IdfParser {
     private static final String INVESTIGATION_TITLE_ID = "Investigation Title";
     private static final String PUBMED_ID = "PubMed ID";
+    private static final String AE_EXPERIMENT_DISPLAY_NAME = "Comment[AEExperimentDisplayName]";
 
     private IdfStreamerFactory idfReaderFactory;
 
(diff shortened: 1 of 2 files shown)


base/src/main/java/uk/ac/ebi/atlas/experimentimport/condensedSdrf/IdfParser.java
```diff
@@ -22,15 +23,26 @@ public IdfParser(IdfStreamerFactory idfReaderFactory) {
...
+                        .orElse("");
+
+            if(title.isEmpty()) {
+                title = titleIdfStreamer.get()
+                                .filter(line -> line.length > 1)
+                                .filter(line -> INVESTIGATION_TITLE_ID.equalsIgnoreCase(line[0].trim()))
+                                .map(line -> line[1])
+                                .findFirst()
+                                .orElse("");
+            }
```
alfonsomunozpomer: You don’t need to set the initial value, `.orElse("")` will take care of that.

Also, I think you can simplify the logic by filtering lines that match either `AE_EXPERIMENT_DISPLAY_NAME` or `INVESTIGATION_TITLE_ID`, no need to go through the stream twice.

alfonsomunozpomer: Much shorter and way DRYer. :)

Notice that you’re doing a case insensitive equals against `AE_EXPERIMENT_DISPLAY_NAME` and `INVESTIGATION_TITLE_ID`, but in the end you’re getting the map values using the original keys. I think that if the fields are all e.g. upper case the code above will fail. `equalsIgnoreCase` is expliciting telling us that the fields might appear in a different case. Either normalize the key to upper/lower case or do regular equals.

Using a map is a very good approach, since we don’t have guarantees about the order of the lines. Again, you can chain the `getOrDefault` methods to the map, but I can see that it may add clarity to do it in two steps.

PR #80, opened 2018-04-19
Bugfix #156697232 - Displayed number assays is 0

monicajianu:
- Created new Cell DataColumnDescriptor
- Displays number of cells analysed rather than number of assays

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/model/experiment/baseline/Cell.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/model/experiment/baseline/Cell.java
@@ -1,0 +1,24 @@
+package uk.ac.ebi.atlas.model.experiment.baseline;
+
+import uk.ac.ebi.atlas.model.BiologicalReplicate;
+import uk.ac.ebi.atlas.model.DescribesDataColumns;
+
+import java.util.Set;
+
+public class Cell extends DescribesDataColumns {
+
+    public Cell(String id) {
+        super(id);
+    }
+
+
+    @Override
+    public Set<String> assaysAnalyzedForThisDataColumn() {
+        return null;
+    }
+
+    @Override
+    public Set<BiologicalReplicate> biologicalReplicatesForThisDataColumn() {
+        return null;
+    }
+}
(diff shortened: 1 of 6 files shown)


base/src/main/java/uk/ac/ebi/atlas/model/experiment/baseline/Cell.java
```diff
@@ -0,0 +1,24 @@
...
+public class Cell extends DescribesDataColumns {
+
+    public Cell(String id) {
+        super(id);
+    }
+
+
+    @Override
+    public Set<String> assaysAnalyzedForThisDataColumn() {
+        return null;
```
alfonsomunozpomer: This, as well as the method below, seems quite dangerous. I strongly suggest returning an immutable empty set. If you want to flag this method as something that shouldn’t be called, throw a non-checked (i.e. wrapped) `OperationNotSupportedException`.

base/src/main/java/uk/ac/ebi/atlas/model/experiment/baseline/SingleCellBaselineExperiment.java
```diff
@@ -0,0 +1,48 @@
...
+                                        ExperimentDisplayDefaults experimentDisplayDefaults) {
+
+        super(type, accession, lastUpdate, displayName, description, disclaimer, species,
+                pubMedIds, experimentDesign, dataProviderURL, dataProviderDescription,
+                alternativeViews, alternativeViewDescriptions, cells, experimentDisplayDefaults);
+    }
+
+    @Override
+    protected JsonObject propertiesForAssay(String runOrAssay) {
+        return null;
```
alfonsomunozpomer: Following the same previous comment, returning [`JsonNull`](https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/JsonNull.html) would be a safer approach.

pcm32:
Could I suggest that we try to add a bit of javadoc at least for classes and public/protected methods. Otherwise is very difficult for new developers to catch up later with the codebase :-).

PR #81, opened 2018-04-20
Bugfix #156874670 - idf file missing in downloads tab

monicajianu:
- Added support for idf file type
- Changed Download tab to display the download links broken down in sections (metadata, results)


sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
```diff
@@ -94,4 +96,16 @@ private JsonObject getExperimentFileAsJson(ExperimentFileType experimentFileType
         return result;
     }
 
+    private JsonObject getDownloadSection(String sectionName, List<ExperimentFileType> experimentFileTypes, String experimentAccession, String accessKey) {
```
alfonsomunozpomer: I applaud your decision of not nameing this method as `getDownloadSectionAsJson`!

monicajianu: Would you like me to remove `asJson` from all the other method names in this class? 😃

PR #83, opened 2018-04-26
Feature/155186331 read cluster files rebased

alfonsomunozpomer:
For some reason when I rebased on `dev` and pushed the branch went back to where it was, and it had some commits prior to the rebase, hence the new branch which sits cleanly on top of `dev`.

Green builds:
• http://bar:8085/browse/ATLAS-BASECIDEVTASKS13-1 (was http://bar:8085/browse/ATLAS-BASECIDEVTASKS5-8)
• http://bar:8085/browse/ATLAS-GXACIDEVTASKS0-1 (was http://bar:8085/browse/ATLAS-GXACIDEVTASKS5-7)
• http://bar:8085/browse/ATLAS-SCXACIDEVTASKS0-1 (was http://bar:8085/browse/ATLAS-SCXACIDEVTASKS5-8)

Goes hand in hand with the PR in https://github.com/ebi-gene-expression-group/scxa-experiment-page-tsne-plot-view with a branch of the same name (minus the “rebased” part).

Some guidance to walk through the branch:
## Reading cluster data
`JsonExperimentTSnePlotController`, the controller previously known as `JsonSingleCellExperimentController`, now has two endpoints, one for cell clusters and another one for gene expression. This controller calls `TSnePlotService` which in turn calls `TSnePlotServiceDao`, which is the class that interfaces with the DB. `TSnePlotServiceDao` reads rows and sends DTOs to the service, which transforms DTOs into problem domain objects (i.e. `TSnePoint`). The DTO is just a bag of values and an inner class of `TSnePoint`. Some functionality of `TSnePlotServiceDao` was previously contained in `TSnePlotDao`.

## Writing cluster data
I simplified the cluster streamers, and removed the factories. They were there for no particular reason and doing something that `SingleCellAnalyticsLoader` could pretty much do on its own. `ClustersStreamer` reads the TSVs and `ClustersDao` inserts the stuff into the DB. Using the DTO class here seemed superfluous. If you think it’s a cleaner design let me know, but this is going to go to production anyway so...

## Other stuff
• I use [AssertJ](https://joel-costigliola.github.io/assertj/) wherever I can. Have a look at the tests and compare it with JUnit/Hamcrest. I’m a big fan of fluent APIs and is very flexible.
• I’ve refactored `experimentimport/analytics/singlecell` in several packages.
• I added `@Transactional` to the loader: we don’t want a half-loaded experiment if something fails (this required changing the loader collaborators to interfaces so that Spring could create the application context in integration tests).


base/src/main/java/uk/ac/ebi/atlas/testutils/RandomDataTestUtils.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Stream;
+
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphabetic;
+import static org.apache.commons.lang3.RandomStringUtils.randomNumeric;
+
+public class RandomDataTestUtils {
+    public static String getRandomExperimentAccession() {
+        // n / 456,975,543,024 chance of clashing for n experiments in the test database, let’s roll!
```
monicajianu: While it is highly unlikely a clash would happen, it would be better if we used a completely clean DB for these tests. This is much easier to implement using Spring Data, because Hibernate handles a lot of the magic for you. It does come with other drawbacks of course.

base/src/main/java/uk/ac/ebi/atlas/testutils/RandomDataTestUtils.java
```diff
@@ -0,0 +1,98 @@
...
+        return "E-" + randomAlphabetic(4).toUpperCase() + "-" + randomNumeric(1, 6);
+    }
+
+    public static String getRandomEnsemblGeneId() {
+        return "ENS" + randomAlphabetic(4).toUpperCase() + randomNumeric(12);
```
monicajianu: Shouldn't this be generating mixed and lower case too? 😉

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -0,0 +1,162 @@
...
+    private final TSnePlotService tSnePlotService;
+
+    @Inject
+    public JsonExperimentTSnePlotController(ScxaExperimentTrader experimentTrader, TSnePlotService tSnePlotService) {
+        super(experimentTrader);
+        this.tSnePlotService = tSnePlotService;
+    }
+
+    @RequestMapping(
+            value = "/json/experiments/{experimentAccession}/tsneplot/{perplexity}/clusters/k/{k}",
```
monicajianu: Is the idea here to have a `/clusters/factors/{factorId}` type of endpoint as well?

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentServiceIT.java
```diff
@@ -101,13 +110,19 @@ public void getValidDownloadsJson() {
...
         assertThat(result.get("suggesterEndpoint").getAsString(), is("json/suggestions"));
 
         assertThat(result.has("ks"), is(true));
-        assertThat(result.get("ks").getAsJsonArray().size(), is(greaterThanOrEqualTo(1)));
+        assertThat(
+                ImmutableSet.copyOf(result.get("ks").getAsJsonArray()).stream()
+                        .map(JsonElement::getAsInt)
+                        .collect(toSet()),
+                containsInAnyOrder(
+                        jdbcTestUtils.fetchKsFromCellClusters(experimentAccession).toArray(new Integer[0])));
```
monicajianu: How come the test is retrieving ks from the database, whereas the `getTsnePlotDataAsJson` method is using the file directly?

atlas-misc/scripts/db_updates/single-cell/20180413-create-clustering-table.sql
```diff
@@ -0,0 +1,11 @@
+DROP TABLE scxa_cell_clusters;
+
+CREATE TABLE IF NOT EXISTS scxa_cell_clusters
+(
+  experiment_accession VARCHAR(255) NOT NULL,
+  cell_id              VARCHAR(255) NOT NULL,
+  k                    INTEGER NOT NULL,
```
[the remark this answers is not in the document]
alfonsomunozpomer: `k` is the number of clusters. A cell can (and will) have a different `cluster_id` depending on the `k` value.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
alfonsomunozpomer: Yes it is. ;)

base/src/main/java/uk/ac/ebi/atlas/experimentimport/analytics/singlecell/SingleCellAnalyticsLoader.java
```diff
@@ -1,48 +1,107 @@
...
+        DataFileHub.SingleCellExperimentFiles files = dataFileHub.getSingleCellExperimentFiles(experimentAccession);
+
+        try (AnalyticsStreamer analyticsStreamer =
+                     new AnalyticsStreamer(files.tpmsMatrix, files.geneIdsTsv, files.cellIdsTsv)) {
+
+            analyticsDao.loadAnalytics(experimentAccession, analyticsStreamer.get());
+        }
+    }
+
+    public void loadTSnePlots(String experimentAccession) {
```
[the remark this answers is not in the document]
alfonsomunozpomer: You’re right. I think `private` is the way to go, as they will never be called directly.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/analytics/singlecell/analytics/Analytics.java
```diff
@@ -0,0 +1,14 @@
+package uk.ac.ebi.atlas.experimentimport.analytics.singlecell.analytics;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Analytics {
```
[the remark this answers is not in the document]
alfonsomunozpomer: I agree but I couldn’t think of anything either. Maybe `CellExpression`?

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentServiceIT.java
alfonsomunozpomer: It indirectly ensures that the clusters file was properly imported to the DB. I also realised it was a bit nefarious, but if at some point it fails, it will signal that something has gone wrong. In the future we want to retrieve the valid ks from the DB as well, and store it in the experiments table (at least I think it’s the best alternative), so we’ll end up getting the ks from there instead of the file.

base/src/main/java/uk/ac/ebi/atlas/testutils/RandomDataTestUtils.java
alfonsomunozpomer: 🤨

PR #89, opened 2018-05-08
Feature/157393774 157393747 157393631 front page improvements

alfonsomunozpomer:
Quick and easy. :)


sc/src/main/webapp/WEB-INF/jsp/tiles/views/home/experiment-list-latest.jsp
```diff
@@ -16,19 +16,9 @@
...
-                                    </c:when>
-                                    <c:otherwise>
-                                        <span data-tooltip class="differential button-rd" title="Differential experiment">D</span>
-                                    </c:otherwise>
-                                </c:choose>
-                            </div>
                             <div class="media-object-section middle hide-for-small-only">
                                 <s:eval expression="T(java.text.NumberFormat).getInstance().format(experimentInfo.numberOfAssays)" var="fmtNumberOfAssays" />
-                                <span class="button secondary no-action">${fmtNumberOfAssays}</span>
+                                <span class="button secondary no-action fixed-width-medium">${fmtNumberOfAssays}</span>
```
monicajianu: A slightly better approach to this would be to use a grid rather than a list. The buttons could take up 20-25% of the width, and the title the rest. Though getting that to be nicely aligned and play nicely would be a massive pain. However it would remove the fixed width, which is generally not ideal.

sc/src/main/webapp/resources/css/atlas.css
```diff
@@ -6,6 +6,10 @@
     font-size: 3em;
 }
 
+#latest-experiments .media-object-section:first-child {
```
monicajianu: A slight annoyance is that the padding at the top and bottom of the latest experiments box are not the same. It makes the list look off-center. It would be nice to disable the padding on the last element in the list to fix this, or make the padding at the top match the one on the bottom and left.

sc/src/main/webapp/WEB-INF/jsp/tiles/views/home/experiment-list-latest.jsp
monicajianu: Also, I think one of the other comments about this was to round the number of cells to something like 10k, 100k, etc. There should be a library that does this (she says naively...).

alfonsomunozpomer: Yes, I agree, but I chose `media-object` becuase In large displays three columns is too wide and two columns is too narrow. Let’s wait for the EBI VF to include [the XY grid](https://foundation.zurb.com/sites/docs/xy-grid.html).

alfonsomunozpomer: We’ll include a code snippet to do this when the time comes. :)

PR #96, opened 2018-05-23
Feature #157167501 - Search endpoint

monicajianu:
- Implemented gene search bundle containing an `ExperimentCard` component used by the `react-faceted-search` component
- Implemented `GeneSearchDao` that retrieves all the cell IDs and all the marker genes associated with a gene ID from Postgres
- Implemented `GeneSearchService` that among others, retrieves facets from Solr given a gene ID
- Added the `json/search/<gene_id>` endpoint to retrieve facets and experiment information given a gene ID
- Added support for the Solr JSON Facet API to the `SolrQueryBuilder` class
- Added retrieval of experiment description from the idf file, to be displayed in the experiment card
- Added integration test for search endpoint
- Minor styling changes and improvements


sc/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -123,26 +123,46 @@ private JsonObject facetValueObject(String group, String value) {
...
         result.addProperty("clusterId", clusterId);
+        result.addProperty("url", url);
 
         return result;
     }
 
-    private String createExperimentPageRelativePath(String contextPath, String experimentAccession) {
-        return String.join("/", contextPath, "experiments", experimentAccession);
+    private String createExperimentPageURL(String experimentAccession, String geneId) {
+       return ServletUriComponentsBuilder.fromCurrentContextPath()
```
alfonsomunozpomer: I was about to write in the commit with `String.join` that there must be libraries to manipulate URIs in a sane way. I was thinking in more general terms, but I like this solution. The fact that it has `fromCurrentContextPath` makes it all the better. 👍

PR #99, opened 2018-05-24
Append ExpressionAtlas gene link at the end of Json being returned to Reach component

haideriqbal:
Not much code edit required for this task just appended the Expression Atlas link at the end of json payload in JsonBioentityInformationController.java
Add new test in JsonBioentityInformationControllerWIT to test if payload contains ExpressionAtlasLink

Links to green builds:
http://bar:8085/browse/ATLAS-GXACIDEVTASKS7
http://bar:8085/browse/ATLAS-BASECIDEVTASKS7-1

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/search/JsonBioentityInformationController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/search/JsonBioentityInformationController.java
@@ -1,5 +1,8 @@
 package uk.ac.ebi.atlas.search;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
 import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -19,6 +22,7 @@ import java.util.stream.Collectors;
 
 import static uk.ac.ebi.atlas.bioentity.properties.BioEntityCardProperties.BIOENTITY_PROPERTY_NAMES;
 import static uk.ac.ebi.atlas.solr.BioentityPropertyName.SYMBOL;
+import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 @RestController
 public class JsonBioentityInformationController extends JsonExceptionHandlingController {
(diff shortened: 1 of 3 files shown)


sc/src/main/java/uk/ac/ebi/atlas/search/JsonBioentityInformationController.java
```diff
@@ -55,6 +59,29 @@ public String getBioentityInformation(@PathVariable String geneId) {
                         geneName,
                         propertyValues);
 
-        return model.get("bioentityProperties").toString();
+
+        // Removed last bracket '[' from the return Json to append the AtlasUrl at the end of this Json
+        String bioentityProperties = model.get("bioentityProperties").toString();
```
alfonsomunozpomer: Patching serialized representations of objects is dangerous and error prone. I strongly discourage you to do this. Start with `JsonArray jsonArray = GSON.fromJson(model.get("bioentityProperties").toString(), JsonArray.class);` and work with `jsonArray`.

haideriqbal: Agreed. Will update the code according to suggestion given.

sc/src/main/java/uk/ac/ebi/atlas/search/JsonBioentityInformationController.java
```diff
@@ -79,9 +76,8 @@ public String getBioentityInformation(@PathVariable String geneId) {
         o.addProperty("name", "ExpressionAtlas");
         o.add("values", values);
 
-        //Added a comma',' and end bracket '[' which was removed earlier to maintain valid Json format
-        String result = bioentityProperties + "," + GSON.toJson(o) + "]";
+        jsonArray.add(o);
 
-        return result;
+        return jsonArray.toString();
```
alfonsomunozpomer: Use `GSON.toJson(jsonArray);`

Even if `toString` does what you’d expect it has the following downsides:
1. Maybe the contract is changed, `toString` could just well be the hash of the object.
2. Gson objects can be built with options to do lenient serialization, pretty printing (with indentations) and other options that aren’t possible with `toString`.

Remember: `Gson.fromJson` reads JSON, `Gson.toJson` writes JSON.

sc/src/test/java/uk/ac/ebi/atlas/search/JsonBioentityInformationControllerWIT.java
```diff
@@ -12,9 +12,7 @@
 import org.springframework.test.web.servlet.setup.MockMvcBuilders;
 import org.springframework.web.context.WebApplicationContext;
 
-import static org.hamcrest.Matchers.greaterThanOrEqualTo;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.isA;
+import static org.hamcrest.Matchers.*;
```
alfonsomunozpomer: It’s good practice to just bring the imports you need. Intent is clearer and it’s impossible you end up having conflicts.

PR #106, opened 2018-05-31
Build information endpoint

monicajianu:
Added `/json/build` endpoint to expose build information such as:
- Bamboo build number
- Git branch of the build
- Git commit ID of the build

The build information is passed from Bamboo as part of the `mvn package` command:
`-Dbuild.number=${bamboo.buildNumber} -Dbuild.branch=${bamboo.planRepository.branch} -Dbuild.revision=${bamboo.planRepository.revision}`

Open to suggestions for a better URL for the endpoint, not sure if `/build` is the most informative.

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/monitoring/BuildVersionController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/monitoring/BuildVersionController.java
@@ -1,0 +1,46 @@
+package uk.ac.ebi.atlas.monitoring;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+import uk.ac.ebi.atlas.controllers.JsonExceptionHandlingController;
+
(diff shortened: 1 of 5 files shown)


sc/src/main/java/uk/ac/ebi/atlas/monitoring/BuildVersionController.java
```diff
@@ -0,0 +1,46 @@
...
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE
+    )
+    public String getBuildNumber() {
+        Map<String, String> buildInformation = new HashMap<>();
+
+        buildInformation.put("Bamboo build version", buildNumber);
+        buildInformation.put("Git branch", buildBranch);
+        buildInformation.put("Git commit ID", buildCommitId);
+
+        return GSON.toJson(buildInformation);
```
alfonsomunozpomer: May be compacted to:
```
        return GSON.toJson(
                ImmutableMap.of(
                        "Bamboo build version", buildNumber,
                        "Git branch", buildBranch,
                        "Git commit ID", buildCommitId));
```

monicajianu:
Codacy failed because it had a rule about test methods needing to contain `assert()`. This obviously doesn't apply for WIT so I've disabled that rule. I wonder if we'll end up disabling most rules... 😃

PR #107, opened 2018-06-04
Bugfix #158005062 - gene search only public experiments

monicajianu:
- Changed gene search queries so only public experiments are retrieved
- Added tests and some test utility methods (e.g. creation of mock single cell experiment)
- Some refactoring


monicajianu:
Bamboo build currently failing for some mysterious reason... http://bar:8085/browse/ATLAS-BASECIDEVTASKS16-GITCO-2/log

monicajianu:
I am also wondering if the Dao class could benefit from more explicit objects, rather than passing maps of maps around...

monicajianu:
1. `getExperimentInformation` is there to retrieve the contents of the `element` field in the JSON response. I did consider putting that block of code in the `GeneSearchController` but then the controller would also need to know about the experiment trader. It is using the `ExperimentAttributesService` as a convenience, it could just as easily build a new JsonObject with properties of the experiment object.
2. Yes, I did think that as well. I am not convinced the DAO is doing too much...  It's just creating objects, which just so happen to be the exact objects the controller needs... I'm not sure there is any value in making different objects, just so you can have additional model conversion layers. I guess what it could do is return Dto objects instead, and have the service create Maps or JsonObjects which can be easily converted to JSON.

PR #110, opened 2018-06-07
Feature #158116264 - Include species facet in search results page

monicajianu:
* Added retrieval of species facet from Solr
* Added name property to Single Cell Schema Fields so that we can have some sort of consistency in naming across the application. For example, curators use the term organism to mean species, and this is stored as `characteristic_organism` in Solr
* Added the species facet group to the list of checkboxes


sc/src/main/java/uk/ac/ebi/atlas/experimentpage/CellMetadataService.java
```diff
@@ -59,7 +60,7 @@ public CellMetadataService(IdfParser idfParser, SolrCloudCollectionProxyFactory
 
         SingleCellAnalyticsSchemaField[] attributeFields = idfParserOutput.getMetadataFieldsOfInterest()
                 .stream()
-                .map(attribute -> SingleCellAnalyticsCollectionProxy.characteristicAsSchemaField(attributeNameToFieldName((attribute))))
+                .map(attribute -> SingleCellAnalyticsCollectionProxy.characteristicAsSchemaField(attributeNameToFieldName((attribute)), attribute))
```
alfonsomunozpomer: We all like Lisp, but those double parenthesis could be removed. 😜

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchServiceDao.java
```diff
@@ -115,10 +115,9 @@ public GeneSearchServiceDao(NamedParameterJdbcTemplate namedParameterJdbcTemplat
                         subFacetValues -> subFacetValues.get("val").toString(),
                         subFacetValues -> subFacetFields
                                 .stream()
-                                .map(SingleCellAnalyticsCollectionProxy.SingleCellAnalyticsSchemaField::name)
                                 .collect(Collectors.toMap(
-                                        subFacetField -> subFacetField,
-                                        subFacetField -> getValuesForFacetField(subFacetValues, subFacetField)
+                                        SingleCellAnalyticsCollectionProxy.SingleCellAnalyticsSchemaField::displayName,
```
alfonsomunozpomer: In these cases, in order to improve readability, I tend to add manually:
```import <package>.SingleCellAnalyticsCollectionProxy.SingleCellAnalyticsSchemaField```

File: sc/src/main/java/uk/ac/ebi/atlas/experimentpage/CellMetadataService.java [hunk 3]
```diff
@@ -96,7 +97,7 @@ public CellMetadataService(IdfParser idfParser, SolrCloudCollectionProxyFactory
         return queryResult.getOrDefault(SingleCellAnalyticsCollectionProxy.FACTORS.name(), Collections.emptyList())
                 .stream()
                 .filter(factor -> !factor.toString().equalsIgnoreCase("single_cell_identifier"))
-                .map(factor -> SingleCellAnalyticsCollectionProxy.factorAsSchemaField(factor.toString()))
+                .map(factor -> SingleCellAnalyticsCollectionProxy.factorAsSchemaField(factor.toString(), StringUtils.capitalize(factor.toString())))
                 .toArray(SingleCellAnalyticsSchemaField[]::new);
```
alfonsomunozpomer: It seems to me that the logic that sets the `displayName` should be encapsulated in `characteristicAsSchemaField` and `factorAsSchemaField`, so that you need to pass a single argument.

monicajianu:
Shouldn't the facet search React component be responsible for not displaying a facet that that only has one value for all results? And surely if we do that with species we should do it with all facets? I'm not sure that is what we want...

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/CellMetadataService.java
monicajianu: `(print "oops!")`

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/CellMetadataService.java
monicajianu: Hmm it might seem like it from the way it's been used, but I think my logic was related to something like the species characteristic. There you wouldn't want the name of the characteristic to be the same as the display name. But seeing as it's not used like this at the moment, I will refactor it.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchServiceDao.java
monicajianu: Done

PR #111, opened 2018-06-11
Feature #158120280: Add Google Analytics to SCEA

monicajianu:
I'm not entirely sure what the difference between [`analytics.js`](https://developers.google.com/analytics/devguides/collection/analyticsjs/) and [`gtag.js`](https://developers.google.com/analytics/devguides/collection/gtagjs/) is... I think `gtag` is a bit more powerful and sends data to services such as Adwords?

### DIFF
--- a/sc/src/main/webapp/WEB-INF/jsp/tiles/base.jsp
+++ b/sc/src/main/webapp/WEB-INF/jsp/tiles/base.jsp
@@ -87,13 +87,22 @@
 
 <!-- JavaScript -->
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
-<script defer="defer" src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js"></script>
+<script defer="defer" src="https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js"></script>
 
 <!-- The Foundation theme JavaScript -->
-<script src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/libraries/foundation-6/js/foundation.js"></script>
-<script src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/foundationExtendEBI.js"></script>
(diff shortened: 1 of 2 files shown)


sc/src/main/webapp/WEB-INF/jsp/tiles/base.jsp
```diff
@@ -87,13 +87,22 @@
...
 <!-- JavaScript -->
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
-<script defer="defer" src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js"></script>
+<script defer="defer" src="https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js"></script>
 
 <!-- The Foundation theme JavaScript -->
-<script src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/libraries/foundation-6/js/foundation.js"></script>
-<script src="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/foundationExtendEBI.js"></script>
+<script src="https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/libraries/foundation-6/js/foundation.js"></script>
+<script src="https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/foundationExtendEBI.js"></script>
```
alfonsomunozpomer: Good catch. I guess I used `dev.ebi.emblstatic.net` when they were rolling out the CDN.

monicajianu: Shall I change in gxa as well?

alfonsomunozpomer:
I hadn’t heard about `gtag.js`. It looks like it’s the New Thing™ to send events to Google Analytics. We’re [tracking some events with `ga`](https://developers.google.com/analytics/devguides/collection/analyticsjs/events) in the heatmap, which looks kind of legacy now. In any case `gtag.js` requires `analytics.js`, so I think it all comes down to whether we want to log events or not. I’ve added a chore not to lose track of this. We might want to bring it up in an upcoming WGC meeting.

sc/src/main/webapp/WEB-INF/jsp/tiles/base.jsp
alfonsomunozpomer: Yes please.

PR #112, opened 2018-06-18
Feature #157944301 - Add health check endpoint for Nagios

monicajianu:
- Implemented new `/json/health` endpoint that indicates whether Solr and the database are up or down
- Implemented `SolrCloudAdminProxy` to handle status check requests via the Solr Admin API (this is not a very generic class at the moment and the name could probably be changed)
- Tests!

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxy.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxy.java
@@ -1,0 +1,49 @@
+package uk.ac.ebi.atlas.solr.cloud.admin;
+
+import org.apache.solr.client.solrj.SolrRequest;
+import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.common.util.NamedList;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component
+public class SolrCloudAdminProxy {
+    private final CloudSolrClient cloudSolrClient;
+
+    public SolrCloudAdminProxy(@Qualifier("zkHost") String zkHost) {
+        cloudSolrClient = new CloudSolrClient.Builder().withZkHost(zkHost).build();
+    }
+
+    // TODO should implement a method that checks multiple collections at the same time
+    public boolean isCollectionUp(String collectionName, boolean isAlias) throws IOException, SolrServerException {
+        SolrRequest request = new CollectionAdminRequest.ClusterStatus();
+
+        final NamedList<Object> response = cloudSolrClient.request(request);
+
+        if(isAlias) {
+            collectionName = getCollectionNameForAlias(response, collectionName);
+        }
+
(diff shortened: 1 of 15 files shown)


monicajianu:
Bamboo green builds:
- http://bar:8085/browse/ATLAS-BASECIDEVTASKS3-5
- http://bar:8085/browse/ATLAS-GXACIDEVTASKS3-5
- http://bar:8085/browse/ATLAS-SCXACIDEVTASKS3-6

base/src/main/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxy.java
```diff
@@ -0,0 +1,81 @@
...
+
+        ArrayList<String> allCollectionNames = new ArrayList<>(collectionNames);
+
+        NamedList<Object> response = cloudSolrClient.request(request);
+
+        // Get real collection names for each alias
+        aliases.forEach(alias -> allCollectionNames.add(getCollectionNameForAlias(response, alias)));
+
+        List<String> statuses = allCollectionNames
+                .stream()
```
alfonsomunozpomer: Instead of the list-wrangling above, I think the following reads better:
```
List<String> statuses = 
        Stream.concat(
            collectionNames,
            Arrays.stream(aliasedCollectionnames).map(alias -> getCollectionNameForAlias(response, alias))
        .flatMap...
```

base/src/main/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxy.java
```diff
@@ -0,0 +1,81 @@
...
+                .collect(Collectors.toList());
+
+        return statuses.isEmpty();
+    }
+
+    // Retrieves the collection name associated with an alias, e.g. the scxa-analytics alias returns scxa-analytics-v2
+    private String getCollectionNameForAlias(NamedList<Object> response, String alias) {
+        LinkedHashMap aliases = (LinkedHashMap) response.findRecursive("cluster", "aliases");
+
+        Object collectionName = aliases.getOrDefault(alias, null);
```
alfonsomunozpomer: `Map::get` will return `null` if the key isn’t found.

base/src/test/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxyIT.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Test
+    public void validCollectionNamesWithAliases() throws IOException, SolrServerException {
+        assertThat(subject.areCollectionsUp(Arrays.asList("bioentities", "analytics"), "scxa-analytics")).isTrue();
+    }
+
+    @Test
+    public void invalidCollectionName() {
+        assertThrows(RuntimeException.class, () -> subject.areCollectionsUp(Collections.singletonList("foo")));
```
alfonsomunozpomer: Remember that [AssertJ can check for exceptions](https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion) too.

monicajianu:
@alfonsomunozpomer The only test class not using Junit 5 is the Mockito one. Looks like the Mockito Junit 5 extension is pretty fresh (https://github.com/mockito/mockito/issues/445), so I'll try to see how well it plays.

base/src/test/java/uk/ac/ebi/atlas/solr/cloud/admin/SolrCloudAdminProxyIT.java
monicajianu: I'll change it to AssertJ, it's much nicer.

PR #113, opened 2018-06-18
Feature/157774516 disable filter options with no results

alfonsomunozpomer:
Companion PR of https://github.com/ebi-gene-expression-group/scxa-faceted-search-results/pull/2.

No links to Bamboo plans are provided since all changes are in the `src/javascript` directory.


sc/src/main/javascript/bundles/gene-search/src/index.js
```diff
@@ -9,7 +9,7 @@ const render = (options, mountNodeId) => {
     <FetchLoader
       host={options.atlasUrl}
       resource={options.resource}
-      ResultElementComponent={ExperimentCard}
+      ResultElementClass={ExperimentCard}
```
monicajianu: Class kind of makes me think of CSS...

PR #114, opened 2018-06-20
Bugfix: IE11 support

monicajianu:


monicajianu:
That commit message should say "to SC" not IE... Microsoft is invading my mind!

PR #118, opened 2018-07-10
Bugfix #158947679 - Clusters.tsv file not closed

monicajianu:
Added try-with-resources statements.


alfonsomunozpomer:
@monicajianu I added a test which may or may not pass... have a look at it, please.

PR #121, opened 2018-07-25
Feature #158116018 - Search genes across species by symbols

alfonsomunozpomer:
### Suggestions:
• Added a new field in the `bioentities` collection, `property_name_id_weight`, that sets an ID precedence based on `property_name` (i.e. in which order suggestions should come from Solr and how a free text query should be searched). This is also encoded in the field `ID_PROPERTY_NAMES` in `BioentitiesCollectionProxy`.
• Added a new suggester without HTML markup that uses the above field as `weighField`.

### Search:
• `SearchController` handles POST and GET requests in different ways: POST expects one `geneQuery` parameter as JSON which is “unfolded” and then forwarded to a GET request which renders the results view. The POST endpoint is the one called by the new [search form component](https://github.com/ebi-gene-expression-group/scxa-gene-search-form).
• `JsonGeneSearchController` now first looks up candidate gene IDs for the searched terms. If none exist an empty `Optional` result is returned. If there are known gene IDs but they can’t be found in SCEA an `Optional` containing an empty set is returned. In both cases the reason is appended in the JSON payload. If there are gene IDs expressed the results are looked up as usual (the logic for serializing into JSON has been quite simplified). This method could be refactored as part of the search service.
• To look up gene IDs in the first phase mentioned above, `GeneIdSearchService` and `GeneIdSearchDao` have been added.
• A new class that embodies the concept of a gene query has been added (`GeneQuery`). At some point we should try to reconcile it with `SemanticQuery` and use the same object.

Many small fixes as I chugged along.

**Green builds**:
• http://bar:8085/browse/ATLAS-BASECIDEVTASKS8-11
• http://bar:8085/browse/ATLAS-GXACIDEVTASKS8-6
• http://bar:8085/browse/ATLAS-SCXACIDEVTASKS8-13


base/src/main/java/uk/ac/ebi/atlas/testutils/SpeciesUtils.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public Species getHuman() {
+        return speciesFactory.create("Homo sapiens");
+    }
+
+    public Species getMouse() {
+        return speciesFactory.create("Mus musculus");
+    }
+
+//    public Species getUnknownSpecies() {
```
monicajianu: Is this commented out block still needed?

base/src/test/java/uk/ac/ebi/atlas/search/suggester/SolrSuggestionReactSelectAdapterTest.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphanumeric;
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.ac.ebi.atlas.solr.cloud.collections.BioentitiesCollectionProxy.ID_PROPERTY_NAMES;
+
+class SolrSuggestionReactSelectAdapterTest {
+    private static final Gson GSON = new Gson();
```
monicajianu: Can we not use the  `GsonProvider` class here?

sc/src/main/java/uk/ac/ebi/atlas/search/SearchController.java
```diff
@@ -1,18 +1,78 @@
...
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static uk.ac.ebi.atlas.solr.cloud.collections.BioentitiesCollectionProxy.ID_PROPERTY_NAMES;
+import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 @Controller
 public class SearchController extends HtmlExceptionHandlingController {
+    private static final List<String> BLAH =
```
monicajianu: Is BLAH an acronym for something? 😉

sc/src/test/java/uk/ac/ebi/atlas/search/GeneSearchServiceDaoIT.java
```diff
@@ -75,39 +74,52 @@ public void validGeneIdReturnKAndClusterIds() {
...
+                subject.getFacets(
+                        emptyList(),
+                        CHARACTERISTIC_INFERRED_CELL_TYPE, CHARACTERISTIC_ORGANISM_PART, CHARACTERISTIC_SPECIES);
 
-        LOGGER.info("Retrieving facets for cell IDs " + String.join(", ", cellIds));
+        assertThat(result).isEmpty();
+    }
 
-        Map<String, Map<String, List<String>>> result = subject.getFacets(cellIds, CHARACTERISTIC_INFERRED_CELL_TYPE, CHARACTERISTIC_ORGANISM_PART, CHARACTERISTIC_SPECIES);
+    private Stream<String> randomGeneIdProvider() {
```
monicajianu: 👍

sc/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
```diff
@@ -13,17 +13,22 @@
 import org.springframework.web.context.WebApplicationContext;
 import uk.ac.ebi.atlas.configuration.WebConfig;
 
+import static java.util.stream.Collectors.toList;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.containsStringIgnoringCase;
+import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.everyItem;
-import static org.hamcrest.Matchers.hasItem;
```
monicajianu: This is removed, but still used in lines 55 and 56...

sc/src/main/java/uk/ac/ebi/atlas/search/SearchController.java
alfonsomunozpomer: D:

base/src/test/java/uk/ac/ebi/atlas/search/suggester/SolrSuggestionReactSelectAdapterTest.java
alfonsomunozpomer: Since it’s a test, I think it’s more correct to use a different instance of `Gson` not to mix responsibilities.

alfonsomunozpomer:
I think the problems you experienced were mostly because the settings in `logback.xml` were overkill and logging coming from Solr queries brought everything to a halt of sorts. I fixed a few of the issues you mentioned though.

PR #124, opened 2018-07-30
Feature #154898174 - Get perplexities from DB

monicajianu:
- Added method to `TsnePlotServiceDao` to retrieve perplexity values given an experiment accession
- Added additional test cases
- Added sorting of perplexity values `TsnePlotViewRoute` component: if the list is not sorted there, the wrong middle value is picked
- Built dev bundles


monicajianu:
@alfonsomunozpomer If we add SQL fixtures, are they integration tests or do they become unit tests...?

PR #125, opened 2018-08-01
Bugfix #159432631 - Experiments with no publication IDs can show wrong paper

monicajianu:
Fixes issue where the string "null" was saved as the Pubmed ID/DOI in the database when an experiment has a publication title, but no PubmedID or DOI. This caused a query to EuropePMC for "null", which for some reason produces a paper about ferrets.

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParserOutput.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParserOutput.java
@@ -3,6 +3,7 @@ package uk.ac.ebi.atlas.experimentimport.idf;
 import uk.ac.ebi.atlas.model.Publication;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -43,7 +44,10 @@ public class IdfParserOutput {
     }
 
     public Set<String> getPubmedIds() {
-        return publications.stream().map(Publication::getPubmedId).collect(Collectors.toSet());
+        return publications.stream()
+                .map(Publication::getPubmedId)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
     }
 
     public List<String> getMetadataFieldsOfInterest() {
@@ -51,6 +55,10 @@ public class IdfParserOutput {
     }
 
     public Set<String> getDois() {
-        return publications.stream().map(Publication::getDoi).collect(Collectors.toSet());
+        return publications
+                .stream()
+                .map(Publication::getDoi)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
     }
 }
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDao.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDao.java
@@ -7,8 +7,10 @@ import uk.ac.ebi.atlas.model.experiment.ExperimentType;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDao.java
```diff
@@ -104,4 +106,12 @@ private ExperimentDTO getSingleExperiment(List<ExperimentDTO> experimentDtos, St
         checkExperimentFound(experimentDtos.size() == 1, accession);
         return experimentDtos.get(0);
     }
+
+    // Returns comma-separated IDs, or null if there are no IDs
+    private String getPublicationIdsAsString(Collection<String> publicationIds) {
+        return publicationIds.stream()
+                .collect(Collectors.collectingAndThen(
+                        Collectors.joining(", "),
+                        (result) -> result.isEmpty() ? null : result));
```
alfonsomunozpomer: Why do you need to return `null` if the resulting string is empty? Have a look at `ScxaExperimentDtoResultSetExtractor` because the logic is now scattered across these two classes.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParserOutput.java
```diff
@@ -43,14 +44,21 @@ public int getExpectedClusters() {
...
         return metadataFieldsOfInterest;
     }
 
     public Set<String> getDois() {
-        return publications.stream().map(Publication::getDoi).collect(Collectors.toSet());
+        return publications
+                .stream()
+                .map(Publication::getDoi)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
```
alfonsomunozpomer: This class has now some behaviour, so a small test would be in order.

sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDao.java
monicajianu: So I would argue that from now on we should add NULL to the database in case we don't have PubMed IDs and I'll change the `ResultSetExtractor` class to check for null or empty string (as existing experiments have empty strings).

base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParserOutput.java
monicajianu: Will do.

sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDao.java
alfonsomunozpomer: I agree that NULL is more correct than an empty string. Sticking to one convention, and a better one, is definitely the best choice.

sc/src/main/java/uk/ac/ebi/atlas/experimentimport/ScxaExperimentDtoResultSetExtractor.java
```diff
@@ -21,13 +21,15 @@ protected ExperimentDTO createExperimentDTO(ResultSet resultSet, String experime
         String accessKeyUUID = resultSet.getString("access_key");
         String title = StringUtils.isEmpty(resultSet.getString("title")) ? "" : resultSet.getString("title");
 
-        String pubMedIdsString = resultSet.getString("pubmed_Ids");
-        Set<String> pubMedIds = StringUtils.isBlank(pubMedIdsString) ?
+        String pubMedIdsString = resultSet.getString("pubmed_ids");
+
+        Set<String> pubMedIds = resultSet.wasNull() || StringUtils.isBlank(pubMedIdsString)?
```
alfonsomunozpomer: I guess that if our DB is clean we should only care about the null case, right?

PR #126, opened 2018-08-01
Feature #157972179 - Changing k or gene ID should unhighlight clusters

monicajianu:
- Added code to reset the highlighted clusters if the colour by or gene boxes change
- Improved code consistency for changing query parameters in experiment page
- FIxed issue where query parameter changes weren't being pushed to the history stack


monicajianu:
I hate IE 😞

monicajianu:
https://www.npmjs.com/package/url-search-params-polyfill I found this... Also, react router's documentation for dealing with multiple query parameters is shockingly bad (i.e. non-existent).

alfonsomunozpomer:
A polyfill is also a good solution (maybe better since we can move to standard stuff). You can package it and add it to the environment as the other polyfills we use. Please add a comment in the code where `URLSearchParams` is used. Some day I hope we can drop support for IE altogether...

PR #135, opened 2018-08-14
Bugfix #159290883 – Average expression in highly expressed gene sets levels results to the same value

alfonsomunozpomer:
For a more in-depth explanation of why this is happening, have a look at [this comment on Pivotal Tracker](https://www.pivotaltracker.com/story/show/159290883/comments/192971214).

The aggregate expression is now averaged over the number of genes expressed in each assay group. This mitigates the issue somewhat, but the underlying problem of using a mean still remains and should be discussed.

I wanted to deploy on www-test to compare with wwwdev, but somehow the automatic deploy didn’t do what I expected, so the new version of the code is in wwwdev. Here’s an example:
• Old version: `http://www-test.ebi.ac.uk/gxa/search?geneQuery=[]&organism=Homo sapiens&conditionQuery=[{"value"%3A"lung"}]`
• New version: `https://wwwdev.ebi.ac.uk/gxa/search?geneQuery=[]&organism=Homo sapiens&conditionQuery=[{"value"%3A"lung"}]`

Besides making small variations more noticeable, it makes average expression levels a bit higher.

I also simplified a few things here and there. There are many questionable design choices in those parts of the code (and I’m guilty of many of them).


monicajianu:
Looks like Bamboo builds are failing for SC: bar:8085/browse/ATLAS-SCXACIDEVTASKS13

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
```diff
@@ -3,18 +3,29 @@
...
 
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.DoubleUnaryOperator;
 
 @Component
 public class BaselineAnalyticsSearchDao {
-    // analyticsQueryClient.baselineFacets searches for baseline experiments
+    static final int MIN_PRECISION = 7;
```
monicajianu: Why magic number 7?

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
```diff
@@ -3,18 +3,29 @@
...
+    static final DoubleUnaryOperator BASELINE_EXPRESSION_ROUNDER =
```
monicajianu: Talking about style guides, isn't all caps snake case more commonly used for constants? I would argue that this rounder is a bit more than a constant...

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
```diff
@@ -3,18 +3,29 @@
...
+            value ->
+                    value >= 1 ?
+                            MathUtils.round(value, 0) :
+                    value >= 0.1 ?
+                            MathUtils.round(value, 1) :
+                    // value < 0.1
+                            MathUtils.round(value, MIN_PRECISION);
+
+    // We don’t need to worry about the type, analyticsQueryClient.baselineFacets filters only baseline experiments
     private static final String EXPERIMENTS_PATH = "$.facets.experimentType.buckets..experimentAccession.buckets[*]";
```
monicajianu: We should try to make this facet retrieving/parsing code a bit more consistent. I had a whole different approach when I implemented `GeneSearchDao`.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
```diff
@@ -50,44 +64,45 @@ public BaselineAnalyticsSearchDao(AnalyticsQueryClient analyticsQueryClient) {
...
 
             @SuppressWarnings("unchecked")
             Map<String, Object> assayGroupIdRoot = (Map<String, Object>) experiment.get("assayGroupId");
             @SuppressWarnings("unchecked")
             List<Map<String, Object>> buckets = (List<Map<String, Object>>) assayGroupIdRoot.get("buckets");
 
-            for(Map<String, Object> assayGroup : buckets)  {
+            for (Map<String, Object> assayGroup : buckets) {
                 String assayGroupId = (String) assayGroup.get("val");
-                double sumExpressionLevel;
```
monicajianu: 👍

base/src/test/java/uk/ac/ebi/atlas/solr/analytics/query/AnalyticsQueryClientIT.java
```diff
@@ -44,127 +51,157 @@
     @Mock
     private RestTemplate restTemplate;
 
+    @Inject
+    private AnalyticsQueryClient goodSubject;
```
monicajianu: Do we need the "good" subject and the (bad) subject?

base/src/test/java/uk/ac/ebi/atlas/solr/analytics/query/AnalyticsQueryClientIT.java
```diff
@@ -44,127 +51,157 @@
...
+        // If we want to go to an extra level of correctness we can get the experiment type from the DB, but if the
+        // type isn’t right in Solr our problems are going to be bigger than this test
+        assertThat(experimentTypes)
+                .isNotEmpty()
+                .allMatch(type -> ExperimentType.get(type).isBaseline());
+    }
 
-        public TestableAnalyticsQueryClient(RestTemplate restTemplate, String solrBaseUrl, Resource baselineFacetsQueryJSON, Resource differentialFacetsQueryJSON,Resource experimentTypesQueryJson, Resource bioentityIdentifiersQueryJson) {
-            super(restTemplate, solrBaseUrl, baselineFacetsQueryJSON, differentialFacetsQueryJSON, experimentTypesQueryJson, bioentityIdentifiersQueryJson);
+    class TestableAnalyticsQueryClient extends AnalyticsQueryClient {
```
monicajianu: Hmm not sure this is such a good idea... AnalyticsQueryClient should be "testable" already! Out of scope for this PR though.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
alfonsomunozpomer: Who knows... this goes back to scientific choices by experts in gene expression. The value was chosen to accomodate proteomics and RNA-Seq experiments under the same baseline category. TPMs lowest value is 0.1 (anything below that we consider it not expressed), but proteomics experiments have [very](https://www.ebi.ac.uk/gxa/experiments/E-PROT-1) [different](https://www.ebi.ac.uk/gxa/experiments/E-PROT-5) magnitudes and measure some sort of fluorescence value IIRC.

base/src/main/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsSearchDao.java
alfonsomunozpomer: `static final` means it’s a constant. Don’t let types deceive you! ;)

base/src/test/java/uk/ac/ebi/atlas/solr/analytics/query/AnalyticsQueryClientIT.java
alfonsomunozpomer: Whatever is being tested with the bad subject (and the `TestableAnalyticsQueryClient`) is a bit pointless. I’m pretty sure that anything worth testing can be achieved with the good subject only, but this was a tangent I didn’t want to take yet. It’s part of our technical debt list, though.

alfonsomunozpomer:
Whatever was failing in Bamboo and Solr before isn’t failing now. https://github.com/ebi-gene-expression-group/atlas/pull/135/commits/401482cc3e9612f925596dc3da5c815b5706579d is *very* annoying, though.

PR #139, opened 2018-08-20
Chore #159855401 - Add Javascript linter

monicajianu:
ESLint setup and file cleanup.


monicajianu:
@alfonsomunozpomer You can ignore this PR for now, I forgot about the gxa bundles! They'll be fun...

alfonsomunozpomer:
Approving because the gxa bundles will be a whole different beast. I think this is all good.

PR #141, opened 2018-08-22
Bugfix #159753830 – Fix ArrayExpress EFO suggestions

alfonsomunozpomer:
The browser was complaining about `$.browser` being `undefined`. It’s an old plugin that the EFO module still uses. Luckily it’s provided by [jQuery Migrate](https://github.com/jquery/jquery-migrate/).


lingyun1010:
The autocomplete is now working fine for Biological Conditions query input.

PR #163, opened 2018-09-11
Bugfix #160415041 - No info retrieved for gene IDs with dot suffixes

monicajianu:
- Implemented fix for gene IDs such as `scaffold_105660.1` were truncated by Spring.
- Added additional test cases.


alfonsomunozpomer:
We had this same issue witht the gene controller at large. We (that is, royal we) missed to fix the gene info too. D:

PR #167, opened 2018-09-25
Bugfix #157046869 - Maintain sdrf column ordering in experiment design

monicajianu:
- Added sdrf parser class that can parse the header of the file and retrieve characteristics and factors, preserving the column ordering
- Added additional logic to the condensed sdrf parser to parse the sdrf (when available), in order to provide the ordered characteristics/factors to the ExperimentDesign class
- Added tests
- Converted `CondensedSdrfParserTest` to AssertJ assertions

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
@@ -96,6 +96,7 @@ public class ExperimentCrud {
 
         UUID accessKeyUuid = accessKey.map(UUID::fromString).orElseGet(UUID::randomUUID);
         experimentDao.addExperiment(experimentDTO, accessKeyUuid);
+
         updateWithNewExperimentDesign(condensedSdrfParserOutput.getExperimentDesign(), experimentDTO);
 
         return accessKeyUuid;
@@ -199,16 +200,10 @@ public class ExperimentCrud {
     }
 
     private void updateWithNewExperimentDesign(ExperimentDesign newDesign, ExperimentDTO experimentDTO) {
-        updateWithNewExperimentDesign(
-                experimentDTO.getExperimentAccession(), experimentDTO.getExperimentType(), newDesign);
-
-    }
-
-    private void updateWithNewExperimentDesign(String accession, ExperimentType type,
-                                               ExperimentDesign experimentDesign) {
         try {
-            experimentDesignFileWriterService.writeExperimentDesignFile(accession, type, experimentDesign);
-            LOGGER.info("updated design for experiment {}", accession);
+            experimentDesignFileWriterService.writeExperimentDesignFile(experimentDTO.getExperimentAccession(),
+                    experimentDTO.getExperimentType(), newDesign);
+            LOGGER.info("updated design for experiment {}", experimentDTO.getExperimentAccession());
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
@@ -1,0 +1,55 @@
+package uk.ac.ebi.atlas.experimentimport.sdrf;
+
+import uk.ac.ebi.atlas.commons.readers.TsvStreamer;
+import uk.ac.ebi.atlas.resource.DataFileHub;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@Named
+public class SdrfParser {
+
+    private final DataFileHub dataFileHub;
+
+    @Inject
+    public SdrfParser(DataFileHub dataFileHub) {
+        this.dataFileHub = dataFileHub;
+    }
+
+    // Returns a map containing the header values for characteristics and factors
+    public Map<String, List<String>> parseHeader(String experimentAccession) {
+        try (TsvStreamer sdrfStreamer = dataFileHub.getExperimentFiles(experimentAccession).sdrf.get()) {
+            Pattern pattern = Pattern.compile("(.*?)(\\[)(.*?)(].*)");
+
+            Map<String, List<String>> headers = new HashMap<>();
+            Arrays.stream(
+                    sdrfStreamer
+                            .get()
+                            .findFirst()
+                            .orElse(new String[0]))
+                    .filter(x -> x.toLowerCase().contains("characteristic") || x.toLowerCase().contains("factor"))
+                    .forEach(header -> {
+                        Matcher matcher = pattern.matcher(header);
+                        if (matcher.matches()) {
+                            String headerType = matcher.group(1).trim().toLowerCase(); // is the header a characteristic or a factor value?
+                            String headerValue = matcher.group(3); // what is the title of the header? e.g. organism, age, etc
+                            if (headers.containsKey(headerType)) {
+                                headers.get(headerType).add(headerValue);
+                            } else {
+                                List<String> headerValues = new ArrayList<>();
+                                headerValues.add(headerValue);
+                                headers.put(headerType, headerValues);
+                            }
+                        }
+                    });
+            return headers;
+        }
+    }
+}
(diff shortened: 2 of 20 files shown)


base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Inject
+    public SdrfParser(DataFileHub dataFileHub) {
+        this.dataFileHub = dataFileHub;
+    }
+
+    // Returns a map containing the header values for characteristics and factors
+    public Map<String, List<String>> parseHeader(String experimentAccession) {
+        try (TsvStreamer sdrfStreamer = dataFileHub.getExperimentFiles(experimentAccession).sdrf.get()) {
+            Pattern pattern = Pattern.compile("(.*?)(\\[)(.*?)(].*)");
```
alfonsomunozpomer, line 35: Funky!

alfonsomunozpomer:
You can remove `SINGLE_CELL_SDRF_FILE_PATH_TEMPLATE` in `DataFileHub` since it’s now part of `ExperimentFiles` and use `SDRF_FILE_PATH_TEMPLATE`.

alfonsomunozpomer (changes requested):
You can remove SINGLE_CELL_SDRF_FILE_PATH_TEMPLATE in DataFileHub since it’s now part of ExperimentFiles and use SDRF_FILE_PATH_TEMPLATE.

I’ll keep going through this tomorrow.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
monicajianu: Perhaps too funky... 😕

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
```diff
@@ -199,16 +200,10 @@ public void updateSingleCellExperimentDesign(String experimentAccession) {
     }
 
     private void updateWithNewExperimentDesign(ExperimentDesign newDesign, ExperimentDTO experimentDTO) {
-        updateWithNewExperimentDesign(
```
alfonsomunozpomer, line 202: 👍 ❗️

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
```diff
@@ -0,0 +1,62 @@
...
+
+            Map<String, Set<String>> headers = new HashMap<>();
+            Arrays.stream(
+                    sdrfStreamer
+                            .get()
+                            .findFirst()
+                            .orElse(new String[0]))
+                    .filter(x -> x.toLowerCase().contains("characteristic") || x.toLowerCase().contains("factor"))
+                    .forEach(header -> {
```
alfonsomunozpomer: This part makes me frown a bit... relying on side effects to populate a map is something I’d in principle discourage. Is there any part here that prevents you from e.g. mapping to a `Pair<String, Set<String>>` and then collect to a map? This case is not especially complex, but having a mapping `Function<>` means that you can also test it on its own if desired.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
```diff
@@ -0,0 +1,62 @@
...
+                            // The header is either a characteristic or a factor value
+                            String headerType = StringUtils.trimAllWhitespace(matcher.group(1)).toLowerCase();
+                            // The title of the header is a type of metadata, such as organism, age, etc
+                            String headerValue = matcher.group(3).trim().toLowerCase();
+                            if (headers.containsKey(headerType)) {
+                                headers.get(headerType).add(headerValue);
+                            } else {
+                                Set<String> headerValues = new LinkedHashSet<>();
+                                headerValues.add(headerValue);
+                                headers.put(headerType, headerValues);
```
alfonsomunozpomer: I’ve seen you ditched Guave elsewhere. I agree that to build new collections it’s not really helpful, but I think it is here:
```
headers.put(headerType, Lists.newArrayList(headerValues));
```

And if you import the static method it reads even better:
```
headers.put(headerType, newArrayList(headerValues));
```

alfonsomunozpomer (approved):
Everything’s good. There are a couple of considerations that I leave to your discretion.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
monicajianu: I got rid of them because their use is discouraged for Java > 7, according to Guava's own javadocs: https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Lists.html#newArrayList--

> Note for Java 7 and later: this method is now unnecessary and should be treated as deprecated. Instead, use the ArrayList constructor directly, taking advantage of the new "diamond" syntax.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
monicajianu: What do you mean by relying on side effects? The fact that a `Map` has unique keys?

monicajianu: I can't see how I can easily map to a `Pair<String, Set<String>`... I can go `String` --> `Pair<String, String>` --> `Map<String, Set<String>>` maybe... Though you won't always get a pair of something (if the matcher doesn't match), so I'd have to filter out the empty pairs... Unless I'm thinking of this in the completely wrong way.

alfonsomunozpomer: By side effects I mean that `forEach` takes a `Consumer` as an argument, which returns `void`, but the map is populated inside.

From [Wikipedia](https://en.wikipedia.org/wiki/Side_effect_(computer_science)):
> In computer science, a function or expression is said to have a side effect if it modifies some state outside its local environment, that is to say has an observable interaction with the outside world besides returning a value.

From [Java 8 API](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html):
> Unlike most other functional interfaces, Consumer is expected to operate via side-effects.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
alfonsomunozpomer: Be aware that it’s discouraged to build an empty modifiable list only. I think somebody liked the factory pattern a bit too much. 😉

base/src/main/java/uk/ac/ebi/atlas/experimentimport/sdrf/SdrfParser.java
alfonsomunozpomer: In essence I can see a function that takes a stream of headers and creates an entry per header if they match a condition. To me that adjusts to a filter-map pattern:
```
String regex = "(.*?)(\\[)(.*?)(].*)";
Pattern pattern = Pattern.compile(regex);
...
Arrays.stream(...)
    .filter(x -> x.toLowerCase().contains("characteristic") || x.toLowerCase().contains("factor"))
    .filter(x -> x.matches(regex))
    .map(x -> // return a pair using the compiled pattern)
    .collect(toMap(Pair::getLeft, Pair::getRight))
```

alfonsomunozpomer: If you rely on the map being partially populated, you can use a `toMap` method with a third argument to merge values: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-java.util.function.BinaryOperator-

alfonsomunozpomer: Another option is to collect using `groupingBy` at the end if you don’t mind ending with a `Map<K, List<V>>`.

monicajianu: That's what I ended up doing because I think it's a bit more suitable... However, it's still not perfect because of the idiosyncrasies of the `Matcher`. You have to call `.matches()` on it in order to get anything useful out of it... And `.matches()` return a boolean, which I guess you could choose to ignore...

alfonsomunozpomer: Beautiful! 😄 In a more serious note, pure functions, wherever possible, are the way to go.

alfonsomunozpomer (approved):

PR #168, opened 2018-09-25
Bugfix - Some experiment titles are wrong

monicajianu:
- Reverted to the old way of retrieving the experiment title from the idf, i.e. retrive the Array Express display name if present, if not retrieve the investigation title

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParser.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentimport/idf/IdfParser.java
@@ -58,23 +58,10 @@ public class IdfParser {
                                             .collect(Collectors.toList()),
                                     (accumulatedValues, newValue) -> accumulatedValues));
 
-            String title;
-            // Experiments that are regularly updated in ArrayExpress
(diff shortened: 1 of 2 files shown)


alfonsomunozpomer:
As a curiosity, I noticed that the condition in [L119 of `IdfParser.java`](https://github.com/ebi-gene-expression-group/atlas/pull/168/commits/7a2cef5869a86a2ce922ae8b37f4f2d2ccb6840f#diff-aa03bd6115b419dd4f28291b225f0c76R119) can be simplified:
```
if (values.isEmpty() || values.stream().allMatch(StringUtils::isBlank)) {
```

Since `allMatch` returns `true` for an empty stream, this can be rewritten as:
```
if (values.stream().allMatch(StringUtils::isBlank)) {
```

Cf.:
```
jshell> Stream.of().allMatch(__ -> false)
$1 ==> true
```
```
jshell> Stream.of(1).allMatch(__ -> false)
$1 ==> false
```

As I said, just a curiosity. I don’t see a bad practice being explicit about `values.isEmpty()` since the semantics of `allMatch` can be somewhat esoteric. :)

alfonsomunozpomer (approved):
Let’s hope we don’t need to re-re-revert this!

PR #170, opened 2018-09-25
Bugfix - Boxplot not showing for GTEx

monicajianu:
... I am not proud of this.


alfonsomunozpomer (approved):
A bug cleanly swept under the carpet... Don’t worry, we’ll fix this!

PR #171, opened 2018-09-25
Chore/160166795 split gxa test data

alfonsomunozpomer:
It’s been a long journey, but at last, here it is...

# Atlas Core 1.0

I think this is a fully revamped version of the module, thus new name and version. It includes a [git submodule](https://github.com/ebi-gene-expression-group/atlas-schemas) and it runs against the new [test data](https://gitlab-internal.ebi.ac.uk/amunoz/atlas-core-test-data), so I’ve set up a new Bamboo plan where it passes:
http://bar:8085/browse/ATLAS-ATLASCORECI0

The way I envision it, once we merge this it would be a good time to split `gxa` and `sc` to their own repositories and we can have modular deployments without branch hell.

If it helps, we can have a look at it together (or parts of it) if you prefer.


pcm32:
Kudos for separate repositories!

monicajianu:
@alfonsomunozpomer It would've been great if the code linting bits of this PR were part of a separate branch. And I'm sure other parts could've been split up in more manageable chunks. As it is, 3209 files have been changed which makes it a little hard to review!

base/src/main/java/uk/ac/ebi/atlas/species/services/PopularSpeciesDao.java
```diff
@@ -21,7 +21,7 @@
             "count(experiment_organism.organism) AS c " +
             "FROM experiment " +
             "LEFT OUTER JOIN  experiment_organism ON experiment_organism.experiment=experiment.accession " +
-            "WHERE private='F' GROUP BY experiment.type, experiment_organism.organism";
+            "WHERE private=FALSE GROUP BY experiment.type, experiment_organism.organism";
```
monicajianu: These changes will require updating every existing experiment in the staging/prod databases, I'm guessing. Something to keep in mind.

monicajianu (changes requested):
What are the steps needed to be done locally so that the application runs properly? As it stands, I cannot run any test or run the applications at all due to various errors popping up everywhere.
I hava already run 
``` 
git submodule init
git submodule update
```

alfonsomunozpomer:
@monicajianu Unfortunately just deleting the test data affects over 3,000 files. I think that the sane way to review this PR is by checking each commit, that’s why I said that you might want to have a look at it with another pair of eyeballs.

Also, yes, the way to get this running is:
```
git submodule init
git submodule update

cd <somewhere_else>
git clone --recurse-submodules git@gitlab-internal.ebi.ac.uk:amunoz/atlas-core-test-data.git
```

The argument `--recurse-submodules` is equivalent to `git clone`, `submodule init` and `submodule update`. You might need to enter your password a few times since the submodules URLs use HTTPS, I noticed this later and will change to SSH but as it is now it’s a bit annoying.

And lastly:
```
ln -s <somewhere_else> <location_of_integration_test_data>
```

You should move your old copy of `integration_test_data`, so that when you switch back to the old branch you don’t need to do all the `ln`s.

monicajianu:
@alfonsomunozpomer It looks like you also need to add `atlas-core` to the sc and gxa `pom.xml` files? Otherwise you get lots of dodgy unknown imports.

alfonsomunozpomer:
Yes. I didn’t mention it because at this stage it’s only this module what needs to be double checked. Interoperability with modules `gxa` and `sc` will be comming soon.

base/src/test/java/uk/ac/ebi/atlas/bioentity/interpro/InterProTSVParserTest.java
```diff
@@ -1,81 +1,69 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
-public class InterProTSVParserTest {
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class InterProTSVParserTest {
     private static final String IPR000001 = "IPR000001";
     private static final String KRINGLE = "Kringle";
```
monicajianu: Apparently a Kringle is a type of Scandinavian pastry! 😄 Methinks a biologist was hungry when they came up with this particular concept.

base/src/test/java/uk/ac/ebi/atlas/model/ArrayDesignTest.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphanumeric;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ArrayDesignTest {
+    @Test
+    void unknownArrayDesignsArePopulatedWithTheirAccession() {
+        assertThat(ArrayDesign.createForUnknownName(randomAlphanumeric(20)))
+                .satisfies(arrayDesign -> assertThat(arrayDesign.accession()).isEqualTo(arrayDesign.name()));
```
monicajianu: While I appreciate that AssertJ has the capability for this `satisfies` thing, I'm not sure it helps readability (or brevity)...

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/api/AssayDesignController.java
```diff
@@ -1,34 +0,0 @@
...
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.ResponseBody;
-import uk.ac.ebi.atlas.model.arraydesign.ArrayDesignDao;
-
-import javax.inject.Inject;
-import java.util.Map;
-
-@Controller
-@Scope("request")
-public class AssayDesignController {
```
monicajianu: What was this used for?

gxa/src/main/java/uk/ac/ebi/atlas/ebeye/DifferentialExperimentContrastsTsvController.java
```diff
@@ -1,79 +0,0 @@
-package uk.ac.ebi.atlas.ebeye;
```
monicajianu: We are still providing some sort of data for ebeye, aren't we? Is that coming straight from data prod?

gxa/src/main/java/uk/ac/ebi/atlas/controllers/rest/api/AssayDesignController.java
alfonsomunozpomer: Maria used it for something on her pipeline. Sorry but I can’t remember the specifics.

gxa/src/main/java/uk/ac/ebi/atlas/ebeye/DifferentialExperimentContrastsTsvController.java
alfonsomunozpomer: There are still two services we provide:
• Endpoint `/das/s4/features`
• The widget

base/src/main/java/uk/ac/ebi/atlas/model/experiment/ExperimentDesign.java
```diff
@@ -150,46 +141,39 @@ public void addAssayHeader(String assayHeader) {
...
-            }
-        }
+    private Map<String, Set<String>> getCharacteristicOntologyTerms() {
+        return samples.entrySet().stream()
+                .collect(toMap(
+                        Map.Entry::getKey,
+                        entry ->
+                                entry.getValue().values().stream()
+                                        .flatMap(sampleCharacteristic ->
+                                                sampleCharacteristic.valueOntologyTerms().stream())
```
monicajianu, line 195: Due to `SampleCharacteristic` using `AutoValue`, it means that the naming convention is ever so slightly different. You have `sampleCharacteristic.valueOntologyTerms()` but you have `factor.*get*ValueOntologyTerms()`. Besides that these getters are pretty much identical... Some inheritance and refactoring the models might simplify things a bit.

monicajianu: Ahh okay, I just noticed the `FactorSet` complication. Nevermind... The whole thing could do with some heavy re-engineering...

base/src/test/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/conditions/ConditionsCollectorTest.java
```diff
@@ -0,0 +1,279 @@
...
+                                    sampleCharacteristics.values().stream()
+                                            .flatMap(headerValueMap -> headerValueMap.values().stream())
+                                            .collect(toSet()))
+                            .build());
+    }
+
+    @Test
+    void differentialConditions() {
+        assertThat(subject.getConditions(differentialExperimentMock))
+//                .isEqualTo(legacyConditionsLookupService.buildPropertiesForDifferentialExperiment(differentialExperimentMock.getAccession(), experimentDesignMock, ImmutableList.of(contrast)))
```
monicajianu: Leftover commented out line

base/src/test/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/conditions/ConditionsCollectorTest.java
```diff
@@ -0,0 +1,279 @@
...
+                                                                .build()
+                                                                .stream())
+                                                .collect(toList()))
+                                .build());
+    }
+
+    private static <K, V> Multimap<K, V> randomizedMultimapOf(Collection<K> buckets, Collection<V> values) {
+        checkArgument(values.size() >= buckets.size());
+
+        List<K> _buckets = Lists.newArrayList(buckets);
```
monicajianu, line 235: What does the underscore in `_variableName` meant to represent...?

alfonsomunozpomer: It’s a way to indicate that it holds a (modified) copy of one of the parameters.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
```diff
@@ -89,11 +84,6 @@ public UUID importExperiment(String experimentAccession, boolean isPrivate) thro
             deleteExperiment(experimentAccession);
         }
 
-        // We only insert in the DB differential and single cell experiments expressions
-        analyticsLoaderFactory
```
monicajianu, line 93: Does data prod have data loading scripts for Expression Atlas yet...?

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
```diff
@@ -122,11 +112,6 @@ public UUID importSingleCellExperiment(String experimentAccession, boolean isPri
...
```
monicajianu, line 127: 👍

base/src/main/java/uk/ac/ebi/atlas/experimentimport/ExperimentCrud.java
alfonsomunozpomer: No, not yet. From the webapp we only need to load coexpressions and index in Solr.

gxa/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageService.java
```diff
@@ -88,7 +93,7 @@ private JsonElement experimentDescription(Experiment experiment,
         return experimentDescription;
     }
 
-    private URI callbackLink(String urlBase, String accessKey, Pair<String, String>... params) {
+    private URI callbackLink(String urlBase, String accessKey, List<Pair<String, String>> params) {
```
monicajianu: Why does this method need to take a list? It is only ever called with a single Pair or no pair. So an `Optional<Pair>`?

base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterService.java
```diff
@@ -22,10 +28,24 @@ public SuggesterService(SuggesterDao suggesterDao, SpeciesFactory speciesFactory
...
-    public Stream<Map<String, String>> fetchGroupedIdSuggestions(String query, String...  species) {
+    public Stream<Map<String, String>> fetchPropertiesWithoutHighlighting(String query, String...  species) {
         Species[] speciesArray = Arrays.stream(species).map(speciesFactory::create).toArray(Species[]::new);
 
         return suggesterDao.fetchBioentityProperties(query, DEFAULT_MAX_NUMBER_OF_SUGGESTIONS, false, speciesArray)
-                .map(suggestion -> ImmutableMap.of("term", suggestion.getTerm(), "category", suggestion.getPayload()));
+                .map(SUGGESTION_TO_MAP);
+    }
+
+    public Stream<Map<String, String>> fetchPropertiesWithHighlighting(String query, String...  species) {
```
monicajianu, line 39: Could you maybe add a comment here saying what the difference between highlighting and no highlighting is? Because there is nothing in the return type that suggests a difference, and it is a very Solr-specific behaviour.

base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterService.java
```diff
@@ -1,18 +1,24 @@
...
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 @Component
 public class SuggesterService {
     // Remember that suggestions are distinct()’ed, so this value is an upper bound
-    private static final int DEFAULT_MAX_NUMBER_OF_SUGGESTIONS = 20;
+    public static final int DEFAULT_MAX_NUMBER_OF_SUGGESTIONS = 20;
+
+    private static final Function<Suggestion, Map<String, String>>
```
monicajianu, line 19: Hmm adding `Suggestion` here couples this service to Solr, when it fact it should be the Dao who is concerned about where the suggestions are coming from.

base/src/test/java/uk/ac/ebi/atlas/model/arraydesign/ArrayDesignDaoIT.java
```diff
@@ -38,7 +43,16 @@
     private ArrayDesignDao subject;
 
     private static final String SELECT_IDENTIFIERS_WITH_DESIGNELEMENT_COUNT =
-            "SELECT identifier, COUNT(DISTINCT(designelement)) FROM designelement_mapping GROUP BY identifier";
+            "SELECT identifier, COUNT(DISTINCT(designelement)) AS count FROM designelement_mapping GROUP BY identifier";
+
+    @BeforeAll
+    void beforeAllTests() {
```
monicajianu: I'm going to be a bit picky here and say that this method name is a bit repetitive and not very descriptive, if you read it together with the annotation. Something like `populateDatabase` would read a bit better IMO.

base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterService.java
alfonsomunozpomer: Sounds sensible.

base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterService.java
alfonsomunozpomer: I agree in principle. The reason why I didn’t do it was that, after looking at [Solr’s `Suggestion` class](https://github.com/apache/lucene-solr/blob/1d85cd783863f75cea133fb9c452302214165a4d/solr/solrj/src/java/org/apache/solr/client/solrj/response/Suggestion.java), it seemed a bit superfluous to create pretty much the same thing under our own namespace. It’s just a data class with no behaviour.

base/src/test/java/uk/ac/ebi/atlas/model/arraydesign/ArrayDesignDaoIT.java
alfonsomunozpomer: You’re totally right.

monicajianu (approved):

alfonsomunozpomer:
@monicajianu Could you please review the last set of changes since your approval?

base/src/main/java/uk/ac/ebi/atlas/configuration/SolrConfig.java
```diff
@@ -15,18 +15,21 @@
...
+        return new HttpSolrClient.Builder("http://" + solrHost + ":" + solrPort + "/solr/analytics").build();
     }
 
     @Bean
-    public CloudSolrClient cloudSolrClient(@Value("${zk.host}") String zkHost) {
-        return new CloudSolrClient.Builder().withZkHost(zkHost + ":2181").build();
+    public CloudSolrClient cloudSolrClient(@Value("${zk.host}") String zkHost,
+                                           @Value("${zk.port}") String zkPort) {
+        return new CloudSolrClient.Builder().withZkHost(zkHost + ":" + zkPort).build();
 //        return SOLR_CLIENT_CACHE.getCloudSolrClient(zkHost);
```
monicajianu: Is commented out line still needed?

alfonsomunozpomer: Removed. Good catch.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/conditions/ConditionsCollector.java
```diff
@@ -36,22 +34,33 @@ public ConditionsCollector(EfoLookupService efoLookupService) {
                         experiment.getExperimentDesign().getSampleCharacteristicsValues(assayAccession).values())
                 .flatMap(Collection::stream)
                 .collect(toSet());
-        }
+    }
 
     // Visits all assays in each factor and builds-collects the conditions in a stream
     public Stream<Condition> getConditions(BaselineExperiment experiment) {
+        ImmutableSetMultimap<String, String> assayAccession2OntologyTerms =
```
monicajianu, line 46: This should probably go in a separate (private) function to avoid code duplication: lines 60-63 seem identical to this block, unless I'm missing something...

pom.xml
```diff
@@ -632,8 +634,13 @@
                 <jdbc.url>jdbc:postgresql:gxpatlasloc</jdbc.url>
                 <jdbc.username>atlas3dev</jdbc.username>
                 <jdbc.password>atlas3dev</jdbc.password>
+                <!-- If you’re running a Beowu^W k8s cluster with minikube, the host of ZK and Solr will be given by
```
monicajianu: Errr... what's a `Beowu^W`?

base/src/main/java/uk/ac/ebi/atlas/testutils/SolrUtils.java
```diff
@@ -1,36 +1,65 @@
...
+//import static uk.ac.ebi.atlas.solr.cloud.collections.AnalyticsCollectionProxy.BIOENTITY_IDENTIFIER;
 
 @Component
 public class SolrUtils {
     private static final int MAX_ROWS = 10000;
     private static final ThreadLocalRandom RNG = ThreadLocalRandom.current();
 
-    private SolrCloudCollectionProxyFactory solrCloudCollectionProxyFactory;
+    private AnalyticsCollectionProxy analyticsCollectionProxy;
+    // private BioentitiesCollectionProxy bioentitiesCollectionProxy;
```
monicajianu: More lonely commented out lines ☹️

pom.xml
alfonsomunozpomer: Ah... [kids these days!](https://en.wikipedia.org/wiki/Talk:Beowulf_cluster#What_Is_Funny_About_/._Jokes?)

base/src/main/java/uk/ac/ebi/atlas/testutils/SolrUtils.java
```diff
@@ -1,36 +1,65 @@
...
+
         return analyticsCollectionProxy
                 .query(queryBuilder)
                 .getResults()
                 .get(RNG.nextInt(MAX_ROWS))
                 .getFieldValue("bioentity_identifier")
                 .toString();
     }
+
+    public String fetchRandomGeneIdFromAnalytics(AnalyticsSchemaField field, String term) {
```
monicajianu, line 42: I'm not sure this method name or its parameters are descriptive enough about what is going on... The method is basically getting a random gene ID that satisfies a condition, but you wouldn't know what without looking at the contents.

pom.xml
monicajianu: 😕

monicajianu:
A couple of very minor things, but on the whole it looks good! Re-approved!

base/src/main/java/uk/ac/ebi/atlas/testutils/SolrUtils.java
alfonsomunozpomer: The fact that you pass a field and a value is a strong enough hint, IMO. Take, for instance, [ThreadLocalRandom.nextInt(int origin, int bound)](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html#nextInt-int-int-); any random generator with arguments imposes certain conditions or restrictions on the returned value. I can add a comment if you want, but I think it’s quite straightforward.

base/src/main/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/conditions/ConditionsCollector.java
alfonsomunozpomer: This is the usual compromise on DRY vs “flow” (for lack of a better word). In this case I thought that adding a level of indirection wouldn’t add value to code readability, but now I’m not so sure. I’ve added a commit to change this. Let me know if you think this reads better.

base/src/main/java/uk/ac/ebi/atlas/testutils/SolrUtils.java
alfonsomunozpomer: Fixed. There was also a method which was no longer used (the no-args random gene ID discussed below).

pom.xml
alfonsomunozpomer: Also: https://en.wikipedia.org/wiki/Backspace#^H

Old jokes from old people. ;)

PR #173, opened 2018-09-28
Feature/160723616 condense marker gene information

lingyun1010:


sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+                                                                    "FROM scxa_marker_genes " +
+                                                                    "WHERE gene_id=:gene_id " +
+                                                                    "GROUP BY experiment_accession) " +
+                                                "AND experiment_accession= :experiment_accession " +
+                                                "AND marker_probability<=:threshold) " +
+                    "OR ((experiment_accession, k) IN ((:experiment_accession, :experiment_and_K)) " +
+                        "AND marker_probability<=:threshold AND gene_id=:gene_id)";
     @Transactional(readOnly = true)
-    public Map<String, Map<Integer, List<Integer>>> fetchKAndClusterIds(String geneId) {
+    public Map<String, Map<Integer, List<Integer>>> fetchKAndClusterIds(String geneId, String experiment_accession, Integer experimentAndK) {
```
monicajianu: Make sure you stick to the naming conventions for variables: they should be `camelCase`, not `snake_case`.

monicajianu: Also, what exactly is an `experimentAndK`? Is it the preferred K value? If yes, the variable name should be descriptive.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+        return namedParameterJdbcTemplate.query(
+                SELECT_PREFFERED_K_STATEMENT,
+                namedParameters,
+                (ResultSet resultSet)->{
+                    Map<String, Map<Integer, List<Integer>>>  result = new HashMap<>();
+
+                    while(resultSet.next()){
+                        String experimentAccession = resultSet.getString("experiment_accession");
+                        IdfParserOutput idfParserOutput = idfParser.parse(experimentAccession);
+                        Integer preferredK = idfParserOutput.getExpectedClusters();
```
monicajianu: The preferred K could also be in the `clusters.tsv` file. The way we chose the preferred k is that we look in the idf file for a k chosen by a curator. If that doesn't exist, we check the `clusters.tsv` file for a k that has the `sel.K` column marked as true. In fact, there is a method that already does this - `TsnePlotSettingsService.getExpectedClusters()`. Try to reuse this if possible.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+    //which is supposed to have the lowest marker_probability,
+    //so we get one pair from searching (experiment_accesion, k).
+    //If dataset has changed, i.e. there are multiple gene clusterID in one preferred cluster K file,
+    //we probably need to order and limit the searching result based on marker_propability
     private static final String SELECT_K_AND_CLUSTER_ID_FOR_GENE_STATEMENT =
-            "SELECT experiment_accession, k, cluster_id FROM scxa_marker_genes AS markers " +
+            "SELECT experiment_accession, k, marker_probability, cluster_id FROM scxa_marker_genes AS markers " +
                     "JOIN scxa_experiment AS experiments ON markers.experiment_accession = experiments.accession " +
-                    "WHERE gene_id=:gene_id AND marker_probability<=:threshold AND private=FALSE";
+                    //"WHERE gene_id=:gene_id AND marker_probability<=:threshold AND private=FALSE " +
```
monicajianu: Remove commented out line if it's no longer required.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
         );
     }
 
+    private static final String SELECT_PREFFERED_K_STATEMENT =
+            "SELECT experiment_accession FROM scxa_marker_genes AS markers "+
```
monicajianu, line 74: Make sure you only retrieve public experiments here.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+    }
+
+    //In marker gene dataset files, one and only one clusterID for each preferred cluster K value,
+    //which is supposed to have the lowest marker_probability,
+    //so we get one pair from searching (experiment_accesion, k).
+    //If dataset has changed, i.e. there are multiple gene clusterID in one preferred cluster K file,
+    //we probably need to order and limit the searching result based on marker_propability
     private static final String SELECT_K_AND_CLUSTER_ID_FOR_GENE_STATEMENT =
-            "SELECT experiment_accession, k, cluster_id FROM scxa_marker_genes AS markers " +
+            "SELECT experiment_accession, k, marker_probability, cluster_id FROM scxa_marker_genes AS markers " +
```
monicajianu, line 105: This SQL could be simplified a lot... 

For example, you don't need to do the join with the `scxa_experiment` table, because you are passing known accessions that you obtained from joining the marker genes table. So you know that every experiment accession you pass in will definitely have that gene ID as a marker gene.

Your `SELECT MIN(marker_probability)` should have the `WHERE experiment_accession = :experiment_accession` because, again, you're only doing this query for a single, known experiment accession.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -71,16 +76,61 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+                        "AND marker_probability<=:threshold AND gene_id=:gene_id)";
     @Transactional(readOnly = true)
-    public Map<String, Map<Integer, List<Integer>>> fetchKAndClusterIds(String geneId) {
+    public Map<String, Map<Integer, List<Integer>>> fetchKAndClusterIds(String geneId, String experiment_accession, Integer experimentAndK) {
         Map<String, Object> namedParameters =
                 ImmutableMap.of(
                         "gene_id", geneId,
-                        "threshold", MARKER_GENE_P_VALUE_THRESHOLD);
+                        "threshold", MARKER_GENE_P_VALUE_THRESHOLD,
+                        "experiment_and_K", experimentAndK,
```
monicajianu: Again, rename `experimentAndK` to something more descriptive.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -92,6 +142,7 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
                         String experimentAccession = resultSet.getString("experiment_accession");
                         Integer k = resultSet.getInt("k");
                         Integer clusterId = resultSet.getInt("cluster_id");
+                        Float marker_probability = resultSet.getFloat("marker_probability");
```
monicajianu: This is not used anywhere. Remove if it's not needed.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -92,6 +142,7 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
 
                         Map<Integer, List<Integer>> kAndClusterIds =
```
monicajianu: You can't really reuse the same code as before for this method, as you have changed its behaviour entirely. This method used to return a map of experiment accessions and their associated ks and cluster IDs. Now, you pass in a _single experiment accession_ and you get ks and cluster IDs. So a lot of this code is now redundant, as it is constructing a map that will only ever have one key (i.e. the accession you passed in as an argument).

monicajianu (changes requested):
While functionally the code is doing its job, quite a few improvements can be made:
- rewrite your Dao methods to remove redundant code
- implement the full "protocol" for retrieved the preferred k, which includes looking at the `clusters.tsv` file
- add some more comprehensive tests, that take into account that your methods now retrive at most 2 marker genes, one with the lowest p-value and one with the preferred k

I am also wondering about the performance of your implementation. Another alternative would be to do a simpler SQL query and then retrieve the lowest p-value and preferred k from the entire list of marker genes. I'm not sure which of the two methods would scale better, but we do have to be careful when doing complex database queries.

alfonsomunozpomer (changes requested):
I think beisdes the points made by @monicajianu there are some OO design issues. The DAO class should only be responsible for communicating with the data store and send some basic representation of the data to the service. Including the `IdfParser` here breaks that. The service should use `TSnePlotSettingsService` to get the preferred `k` and send it to the DAO, which will only handle communication with the DB.

Also, refrain from changing the API and names of existing methods. Add your own and then remove the old ones if they’re not used.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
lingyun1010: As far as I know, this SQL cannot be simplified anymore, since we have two filter schema and for each accession, we may have one or two return entries.

I have tried what you suggested anyway, but it failed to retrieve the results with two entries.

monicajianu: I tried this query straight on the database and it does return 2 rows... 
```
SELECT experiment_accession, k, marker_probability, cluster_id FROM scxa_marker_genes AS markers
WHERE (marker_probability IN (
                               SELECT MIN(marker_probability)
                               FROM scxa_marker_genes
                               WHERE gene_id='ENSG00000013725'
                                 AND experiment_accession='E-GEOD-70580'
                               GROUP BY experiment_accession
                               )
                 AND marker_probability <= 0.05)
     OR ((experiment_accession, k) IN (
      ('E-GEOD-70580', 4))
           AND marker_probability<=0.05
           AND gene_id='ENSG00000013725');
```

In this case, k = 4 is the preferred K value for E-GEOD-70580 (coming from the idf file) and it's not the same as the k for the gene with the minimum p-value.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
lingyun1010: I have tried to use `TsnePlotSettingsService.getExpectedClusters()`, unfortunately, I always got an error "There was a problem processing your request. Please try again later." We can have a look together.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
lingyun1010: I do not quite know how to solve this. Is it related to `private` attribute in the dataset?

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -2,23 +2,21 @@
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.gson.JsonObject;
```
monicajianu: Remove unused import.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -31,13 +29,16 @@
     private static final double MARKER_GENE_P_VALUE_THRESHOLD = 0.005;
 
     private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+    private final TsnePlotSettingsService tsnePlotSettingsService;
```
monicajianu: This is no longer being used. Please remove.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -65,49 +66,80 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+        );
+    }
+
+    private static final String SELECT_EXPERIMENT_ACCESSION_FOR_GENE_ID =
+            "SELECT experiment_accession FROM scxa_marker_genes AS markers "+
+            "JOIN scxa_experiment AS experiments ON markers.experiment_accession = experiments.accession "+
+            "WHERE private=FALSE AND gene_id=:gene_id "+
+            "GROUP BY experiment_accession";
+    @Transactional(readOnly = true)
+    public List<String> preferredK (String geneId) {
```
monicajianu: Why does this method have the name `preferredK` when it actually retrieves a list of experiment accessions? The method name should describe its behaviour. That way, I don't have to look at the method body to know what it is doing.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -65,49 +66,80 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
+                                                                    "FROM scxa_marker_genes " +
+                                                                    "WHERE gene_id=:gene_id " +
+                                                                    "GROUP BY experiment_accession) " +
+                                                "AND experiment_accession= :experiment_accession " +
+                                                "AND marker_probability<=:threshold) " +
+                    "OR ((experiment_accession, k) IN ((:experiment_accession, :preferred_K)) " +
+                        "AND marker_probability<=:threshold AND gene_id=:gene_id)";
     @Transactional(readOnly = true)
-    public Map<String, Map<Integer, List<Integer>>> fetchKAndClusterIds(String geneId) {
+    public  Map<Integer, List<Integer>> fetchPreferredKAndMinPAndClusterIds(String geneId, String experimentAccession, Integer preferredK) {
```
monicajianu: This method name could be improved further, I feel. If I read it now, I understand that it is retrieving the preferred K value, the minimum probability and some cluster IDs. That is not quite what it is doing. It is retriving _cluster IDs for_ the preferred K and the K with the minimum probability. 
Naming method is difficult, especially if the behaviour is complex, but we shouldn't be misleading in our naming.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -3,10 +3,10 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import org.springframework.stereotype.Component;
+import uk.ac.ebi.atlas.experimentpage.TsnePlotSettingsService;
 
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.sql.ResultSet;
```
monicajianu: Remove unused import.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -3,10 +3,10 @@
...
+import java.util.*;
```
monicajianu: IntelliJ tries to be clever here about grouping imports, but it's not very good practice to use `import *`. To fix this, go to IntelliJ IDEA --> Preferences --> Editor --> Java --> Imports tab and change the number in the box (see screenshot below).

screen shot 2018-10-19 at 10 06 43

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -66,4 +69,22 @@ public GeneSearchService(GeneSearchDao geneSearchDao) {
             forkJoinPool.shutdown();
         }
     }
+
+    private  Map<String, Map<Integer, List<Integer>>> fetchPreferredKAndMinP (List<String> experimentAccessions, String geneId){
```
monicajianu: Again, this method name is not quite right (I'm sorry!). Is it just returning the preferred K and the minimum probability?

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -66,4 +69,22 @@ public GeneSearchService(GeneSearchDao geneSearchDao) {
...
+        Map<String, Map<Integer, List<Integer>>> result = new HashMap<>();
+
+        for (String experimentAccession : experimentAccessions) {
+            Optional<Integer> optional = tsnePlotSettingsService.getExpectedClusters(experimentAccession);
```
monicajianu: What does this variable represent? The name should reflect it.

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -66,4 +69,22 @@ public GeneSearchService(GeneSearchDao geneSearchDao) {
...
+            if (optional.isPresent()) {
+
+                Integer value = optional.get();
```
monicajianu: An assignment is not strictly necessary. You can just do the `.get()` when you pass the parameter below.

monicajianu (commented):
Getting there! I think the separation of concerns is good, every class does what it's supposed to. The only thing left is to improve method naming and testing. To help find code that hasn't been covered by tests, run the tests with coverage in IntelliJ. It should tell you which lines are/aren't covered. We should aim for as close to 100% coverage as possible.

sc/src/test/java/uk/ac/ebi/atlas/search/GeneSearchServiceTest.java
```diff
@@ -9,15 +9,18 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockito.junit.jupiter.MockitoSettings;
 import org.mockito.quality.Strictness;
+import uk.ac.ebi.atlas.experimentimport.admin.Op;
```
monicajianu: Unused import here. To prevent these things from happening to you go to Code --> Optimise imports in Intellij every once in a while when developing. That should get rid of any unused imports.

monicajianu (approved):
Looks good! Great job!

alfonsomunozpomer (changes requested):
I don’t have much to add. Very interesting conversation and seeing @lingyun1010’s code getting better with each revision. Good job both of you!

The only last thing I’d like to comment is that in `GeneSearchServiceTest` you’re using hard-coded values several times (to stub the methods first, and to check the expected results later). Instead of that, declare a static constant and reuse it. Also, instead of `ENSFOOBAR1` or `E-MTAB-0000` use the random generators from `RandomDataTestUtils`, so that the name changes with each test run and tests can capture better the invariants in your code.

alfonsomunozpomer (approved):
Good stuff!

PR #174, opened 2018-10-09
Feature #160591968 - Additional attributes in dropdown

monicajianu:
- Added additional attributes from the IDF to the experiment page JSON payload, such that they can be displayed in the colour by dropdown. 
- Added tests.

### DIFF
--- a/sc/src/test/java/uk/ac/ebi/atlas/metadata/CellMetadataDaoIT.java
+++ b/sc/src/test/java/uk/ac/ebi/atlas/metadata/CellMetadataDaoIT.java
@@ -13,6 +13,7 @@ import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.test.context.web.WebAppConfiguration;
 import uk.ac.ebi.atlas.configuration.WebConfig;
+import uk.ac.ebi.atlas.experimentimport.idf.IdfParser;
 import uk.ac.ebi.atlas.solr.cloud.SolrCloudCollectionProxyFactory;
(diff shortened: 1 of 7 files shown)


sc/src/test/java/uk/ac/ebi/atlas/metadata/CellMetadataDaoIT.java
```diff
@@ -151,6 +167,19 @@ void getFieldValuesForNoFieldsReturnsEmpty() {
...
+    private Stream<String> experimentsWithAdditionalAttributesProvider() {
+        return jdbcUtils.getPublicSingleCellExperimentAccessions()
+                .stream()
+                .filter(accession -> !idfParser.parse(accession).getMetadataFieldsOfInterest().isEmpty());
+    }
+
+    private Stream<String> experimentsWithoutAdditionalAttributesProvider() {
+        return jdbcUtils.getPublicSingleCellExperimentAccessions()
+                .stream()
+                .filter(accession -> idfParser.parse(accession).getMetadataFieldsOfInterest().isEmpty());
```
alfonsomunozpomer: Notice that for both choosing the experiments with or without additional metadatafields, and for checking the results in the tests, you’re following the same code path: `idfParser.parse(experimentAccession).getMetadataFieldsOfInterest()`.

In order to alleviate that I think it would be better to inject `DataFileHub` and scan the line with additional attributes, so that the experiment-picking logic is more isolated from the implementation of the test subject.

alfonsomunozpomer (approved):
You could improve the test a bit marginally by having a different piece of logic to choose experiments. Other than that everything looks good.

PR #175, opened 2018-10-09
Feature #160591849 - Add additional attributes to tooltip

monicajianu:
Note: I branched from https://github.com/ebi-gene-expression-group/atlas/tree/feature/160591968-add-additional-attributes-to-dropdown so you can ignore the first commit. There is an [additional PR](https://github.com/ebi-gene-expression-group/atlas/pull/174) for that.

- Added additional attributes to the metadata retrived through the cell metadata endpoint
- Cleaned up code


alfonsomunozpomer (approved):
LGTM.

PR #177, opened 2018-10-14
Use H2 DB and new test data in ITs in gxa module

alfonsomunozpomer:
The journey continues! The most significant changes were in the base module. This is quite simple by comparison.

I created [an alternative plan in Bamboo](http://bar.ebi.ac.uk:8085/browse/ATLAS-GXACI0) for this version of the module since it uses a different Solr instance and test data directory. [I moved the test data repo to the regular (non-internal) EBI’s GitLab](https://gitlab.ebi.ac.uk/amunoz/atlas-test-data).

The PR is done against the original branch, so we can rebase onto it and then merge onto dev.


gxa/src/test/java/uk/ac/ebi/atlas/experimentimport/ExperimentCheckerIT.java
```diff
@@ -1,46 +1,79 @@
...
+        populator.addScripts(new ClassPathResource("fixtures/experiment-delete.sql"));
+        populator.execute(dataSource);
     }
 
-    @Test
-    public void testCheckDifferential() {
-        subject.checkAllFiles(RNA_SEQ_DIFFERENTIAL_EXPERIMENT_ACCESION, ExperimentType.RNASEQ_MRNA_DIFFERENTIAL);
+    @ParameterizedTest
+    @MethodSource("experimentAccesionWithTypeProvider")
+    void testCheckBaseline(String experimentAccession, ExperimentType experimentType) {
```
monicajianu: This method seems to be checking baseline and differential which doesn't match the method name.

gxa/src/test/java/uk/ac/ebi/atlas/experimentimport/ExperimentCheckerIT.java
```diff
@@ -1,46 +1,79 @@
...
-public class ExperimentCheckerIT {
-    private static final String RNA_SEQ_BASELINE_EXPERIMENT_ACCESSION = "E-MTAB-513";
-    private static final String PROTEOMICS_BASELINE_EXPERIMENT_ACCESSION = "E-PROT-1";
-    private static final String RNA_SEQ_DIFFERENTIAL_EXPERIMENT_ACCESION = "E-MTAB-4289";
-    private static final String MICROARRAY_1_COLOUR_DIFFERENTIAL_EXPERIMENT_ACCESION = "E-GEOD-34130";
-    private static final String MICROARRAY_2_COLOUR_DIFFERENTIAL_EXPERIMENT_ACCESION = "E-GEOD-43049";
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = TestConfig.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class ExperimentCheckerIT {
```
monicajianu, line 34: Could do with a negative test case in this class, to make sure the `ExperimentChecker` does blow up when confronted with an unknown experiment.

gxa/src/test/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/SolrInputDocumentInputStreamIT.java
```diff
@@ -91,14 +114,15 @@ public void testSomeExperiments() throws Exception {
         assertThat(count, is(greaterThan(100)));
```
monicajianu: I realise this is an older bit of code and out of scope for this PR but this assertion is a bit odd... Why 100? Also, I don't get the point of the `count` variable. Why not just do `result.size()` instead?

gxa/src/test/java/uk/ac/ebi/atlas/experimentpage/json/opentargets/EvidenceServiceIT.java
```diff
@@ -43,23 +53,38 @@
...
-        MicroarrayExperiment experiment = (MicroarrayExperiment) experimentTrader.getPublicExperiment("E-GEOD-12685");
-        subject.evidenceForExperiment(experiment, contrast -> {
-            MicroarrayRequestPreferences requestPreferences = new MicroarrayRequestPreferences();
-            requestPreferences.setFoldChangeCutoff(0.0);
-            requestPreferences.setCutoff(1.0);
-            requestPreferences.setHeatmapMatrixSize(1000);
-            requestPreferences.setSelectedColumnIds(ImmutableSet.of(contrast.getId()));
-            return new MicroarrayRequestContext(requestPreferences, experiment);
-        }, listBuilder::add);
+        MicroarrayExperiment experiment = (MicroarrayExperiment) experimentTrader.getPublicExperiment("E-GEOD-43049");
```
monicajianu, line 78: It looks like this test escaped parameterization and it still has hard-coded experiment accessions. Unless it's changed in a later commit?

gxa/src/test/java/uk/ac/ebi/atlas/experimentimport/ExperimentCheckerIT.java
alfonsomunozpomer: Yes, there are other ways to improve this test (such as having experiments where individual files are missing), but this falls out of the current PR.

gxa/src/test/java/uk/ac/ebi/atlas/experimentimport/analyticsindex/SolrInputDocumentInputStreamIT.java
alfonsomunozpomer: I’ll change it to 0.

As for your second question, `result` extends `Iterable` and there’s no way to get the size of the result other than by counting.

gxa/src/test/java/uk/ac/ebi/atlas/resource/ContrastImageControllerWIT.java
```diff
@@ -41,54 +50,72 @@
...
     @BeforeEach
     void setUp() {
         this.mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
     }
 
     @Test
-    void responseForGseaPlotsShouldBeNonEmpty() throws Exception {
+    void responseForGseaPlotsShouldNotBeEmpty() throws Exception {
         responseAssertions(
-                this.mockMvc.perform(get(MessageFormat.format(GSEA_PLOT_IMAGE_URL_TEMPLATE, "g6_g2", "go"))));
```
monicajianu, line 56: Interesting that you chose to stick to specific experiment IDs here. When I write WITs with random experiments, I just end up checking the JSON structure is okay, rather than the content. I do think we should check for content, like you do here. But perhaps eventually we would have files with expected JSON responses for each experiment in the fixtures? I'm not sure what the best way of doing it would be.

gxa/src/test/java/uk/ac/ebi/atlas/tracks/TracksControllerWIT.java
```diff
@@ -1,84 +1,113 @@
...
         this.mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
     }
 
     @Test
-    public void getTrackFile() throws Exception {
-        this.mockMvc.perform(get(URL_TEMPLATE.format(new Object[] {"E-MTAB-513", "g1"})))
+    void getTrackFile() throws Exception {
+        this.mockMvc.perform(get(BASELINE_URL_TEMPLATE.format(new Object[] {"E-MTAB-5423", "g391"})))
                 .andExpect(status().isOk())
                 .andReturn();
```
monicajianu: Again, probably out of scope for this PR, but I don't think these `andReturn()` statements are needed.

gxa/src/test/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsFacetsReaderTest.java
```diff
@@ -14,7 +14,161 @@
 import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 public class BaselineAnalyticsFacetsReaderTest {
-    private static final JsonObject RESPONSE_2_JSON_FACETS = GSON.fromJson(
+    private static final String JSON_RESPONSE =
```
monicajianu, line 17: Any reason you decided to inline this? It looks a bit messy, it's probably easier to read as a proper JSON file.

monicajianu (approved):
This is great stuff! Approved!

gxa/src/test/java/uk/ac/ebi/atlas/resource/ContrastImageControllerWIT.java
alfonsomunozpomer: This is from a time when we were writing worse tests. Some of them I parameterized, others I just changed the hard-coded actual/expected values. In principle yes, the best thing would be to test structure or that the values match a set of possible values.

gxa/src/test/java/uk/ac/ebi/atlas/tracks/TracksControllerWIT.java
alfonsomunozpomer: I removed them, while we’re at it.

gxa/src/test/java/uk/ac/ebi/atlas/solr/analytics/baseline/BaselineAnalyticsFacetsReaderTest.java
alfonsomunozpomer: I think that in this case, because it’s only used once, inlining puts it closer to the actual use. If you look a the resources you need to look around a bit more to know who is responsible for keeping that file. For instance, if the test is removed or changed there’s some risk of cruft being left around.

PR #178, opened 2018-10-14
Use H2 DB and new test data in ITs in sc module 

alfonsomunozpomer:
The last stop in a journey of wonder and excitement.

As before, I’ve added [a new plan in Bamboo](http://bar.ebi.ac.uk:8085/browse/ATLAS-SCXACI0-3). Test experiments are now only two. It didn’t seem that adding any others would add to the spread of features, but maybe I’m missing something.

Have a look at the script to create fixtures. It’s funny that something as rudimentary and ugly works.


sc/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
```diff
@@ -45,6 +53,26 @@
 
     private GeneIdSearchDao subject;
 
+    @BeforeAll
+    void populateDatabaseTables() {
+        ResourceDatabasePopulator populator = new ResourceDatabasePopulator();
+        populator.addScripts(
+//                new ClassPathResource("fixtures/scxa_experiment-fixture.sql"),
```
monicajianu:
```suggestion
```

monicajianu: Apparently you can suggest specific code changes which you can commit straight away? It's some new fancy feature.

sc/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
```diff
@@ -45,6 +53,26 @@
...
+                new ClassPathResource("fixtures/scxa_analytics-fixture.sql"));
+//                new ClassPathResource("fixtures/scxa_marker_genes-fixture.sql"));
```
monicajianu:
```suggestion
```

monicajianu: Although in this case it's a bit silly because you'd end up with 4 separate commits for each changed line, I think. Unless you can do it in bulk but I can't figure out how...

monicajianu (approved):
Looks good!

PR #180, opened 2018-10-17
Feature/160724655 show GitHub page release notes

lingyun1010:


alfonsomunozpomer (changes requested):
Nice work! It’s really great to see this working. :)

The only downside, compared to our current page, is that the iframe height isn’t dynamically adjusted, and we end up with two scroll bars. Could you please have a look at [Iframe-resizer](https://github.com/davidjbradshaw/iframe-resizer)? Notice that you need to add a `<script>` tag in both the GitHub page and our `release-notes.html`.

alfonsomunozpomer (approved):

PR #182, opened 2018-10-24
Sanitize JS bundles

alfonsomunozpomer:
It’s embarrassing to realize that the bundles were successfully built because:
• They were using an old version of React
• In the case of the Differential Expression component because the transpiled JS was being checked in

A bit of maintenance so that when release time comes this doesn’t catch us off-guard.

P.S.: Feedback Module R.I.P. (we’ll do a better one)


monicajianu (approved):
👍 Good riddance, smiley faces! Should we obliterate `/javascript/atlas_modules/feedback` as well or are we keeping it for entertainment purposes?

PR #183, opened 2018-10-25
Chore #160166795 – Post merge maintenance

alfonsomunozpomer:
Some odd loose ends that were still dangling from the Big Merge™.

I also took the opportunity to bump the artifact versions. Since there are not going to be any new releases before the one scheduled for November it’s safe to bump gxa and sc to `<new_version>-SNAPSHOT`.

@lingyun1010 Please have a look at this one, too.


monicajianu (approved):
Looks good!

pom.xml
```diff
@@ -611,23 +611,6 @@
     <profiles>
         <profile>
             <id>dev</id>
-            <properties>
```
lingyun1010, line 614: By removing the old dev properties means we do not use the local database anymore, is it? I am wondering if keeping both of dev and dev.local is good in case local database server breaks somehow.

lingyun1010 (commented):
Looks good for me too. Just a suggestion on keeping both <properties> set ups.

pom.xml
alfonsomunozpomer: The local (Docker or native) database is the one we’re keeping. If you mean `gxpatlasdev` on `pgsql-hxvm-037.ebi.ac.uk`, that one’s going away soon. It reduces the number of environments we need to keep track of and is a step towards containerizing the whole application.

PR #184, opened 2018-10-26
Chore - Build SC bundles

monicajianu:
- Updated all bundles to Webpack 4
- Fixes small issue with default selection in colour by dropdown
- Built dev bundles

As it is, these bundles don't include the latest version of the scxa-experiment-page-tsne-plot-view component, [found in this PR](https://github.com/ebi-gene-expression-group/scxa-experiment-page-tsne-plot-view/pull/17). After those changes are merged and a new package published, the bundles should be rebuilt and merged. Otherwise we'll have a broken tooltip in dev!


alfonsomunozpomer (approved):
Now with more Webpack. 😃

PR #185, opened 2018-10-26
Feature #161088610 - Implement endpoint for experiments by species

monicajianu:
- Implemented popular species endpoint for SC
- Implemented CardModel and factory method to produce cards that implement the [agreed schema](https://www.ebi.ac.uk/seqdb/confluence/display/GXA/Home+page+card+JSON+schema)
- Lots of refactoring and extracting of common functionality
- Small UI changes in SC - replaced organism with species and removed Comparisons column in the experiments table
- Removed rounded corners in Expression Atlas
- Testing


base/src/main/java/uk/ac/ebi/atlas/home/PopularSpeciesService.java
```diff
@@ -1,45 +1,44 @@
...
+                popularSpeciesInfos.get().stream()
                         .filter(speciesInfo -> kingdom.equalsIgnoreCase(speciesInfo.kingdom()))
                         .collect(Collectors.toList());
 
-        return filteredList.subList(0, Math.min(filteredList.size(), howMany));
+        return getSublist(filteredList, howMany);
     }
 
+    private List<PopularSpeciesInfo> getSublist(List<PopularSpeciesInfo> list, int length) {
+        return list.subList(0, Math.min(list.size(), length));
```
alfonsomunozpomer, line 42: It’s a bit stupid that Java doesn’t have this yet (I’ve typed this quite a few times). I guess that now that streams are all the rage it will never get implemented since you can achieve this with `.stream().limit(length)`. Up to you.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.List;
+import java.util.Optional;
+
+public class CardModelAdapter {
+
```
alfonsomunozpomer, line 11: Remember that any `JsonObject` can be backed by a `Map` and any `JsonArray` by a `List`. Gson will do the rest for you. I think you can simplify a bit of code here and rely less on side-effects (i.e. `forEach`).

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModel.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.google.auto.value.AutoValue;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.List;
+import java.util.Optional;
+
+@AutoValue
+public abstract class CardModel {
+    public abstract String iconType();
```
alfonsomunozpomer: Maybe `iconType` should be an enum?

base/src/test/java/uk/ac/ebi/atlas/model/card/CardModelAdapterTest.java
```diff
@@ -0,0 +1,129 @@
...
+        CardModel cardModel = getRandomCardModel(true, 3, false);
+
+        JsonObject result = CardModelAdapter.serialize(cardModel);
+
+        assertThat(result.has("iconDescription")).isTrue();
+
+        JsonArray contentResult = result.get("content").getAsJsonArray();
+        assertThat(contentResult.size()).isEqualTo(3);
+
+        contentResult.forEach(x -> {
```
alfonsomunozpomer, line 53: You really don’t like `allSatisfy`...

alfonsomunozpomer (commented):
It all looks good. My only two observations that would rather discuss is making `iconType` an enum instead of a string. I don’t know if there was any rationale behind that choice. Also I suspect that `CardModelAdapter` may be simplified.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
monicajianu: Having the `serialize` method just return a string would make it extremely difficult to test. And if it returns `Map<String, Object>`, then it shouldn't really be called serialize anymore. Or last option, it still returns a `JsonObject` but then the implementation becomes overly convoluted and you start mixing these `Maps` and `JsonObjects` everywhere and it all gets confusing...

PR #191, opened 2018-11-01
Feature/161418158 move number cells highcharts

lingyun1010:
Add a method in TsneServiceDao to fetch the `numberOfCells` by `experimentAccession`. Save the attribute value in `model`. Display a line to show the `numberOfCells` in the `experiment-description`.

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentController.java
@@ -48,8 +48,10 @@ public class ExperimentController extends HtmlExceptionHandlingController {
         Experiment<Cell> experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
 
         model.addAllAttributes(experimentAttributesService.getAttributes(experiment));
-        model.addAttribute("content", GSON.toJson(experimentPageContentForExperiment(experiment, accessKey)));
-        model.addAttribute("numberOfCells", tSnePlotServiceDao.fetchCellNumberByExperimentAccession(experimentAccession));
+        JsonObject content = experimentPageContentForExperiment(experiment, accessKey);
+        content.addProperty("numberOfCells", tSnePlotServiceDao.fetchNumberOfCellsByExperimentAccession(experimentAccession));
+        model.addAttribute("content", GSON.toJson(content));
+        model.addAttribute("numberOfCells", content.get("numberOfCells"));
 
         return "experiment-page";
     }
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentController.java
@@ -14,6 +14,7 @@ import uk.ac.ebi.atlas.model.experiment.ExperimentDesignTable;
 import uk.ac.ebi.atlas.model.experiment.baseline.Cell;
 import uk.ac.ebi.atlas.resource.DataFileHub;
 import uk.ac.ebi.atlas.trader.ScxaExperimentTrader;
+import uk.ac.ebi.atlas.tsne.TSnePlotServiceDao;
 
 import javax.inject.Inject;
 
@@ -25,15 +26,18 @@ public class ExperimentController extends HtmlExceptionHandlingController {
     private final DataFileHub dataFileHub;
     private final ExperimentPageContentService experimentPageContentService;
     private final ExperimentAttributesService experimentAttributesService;
+    private final TSnePlotServiceDao tSnePlotServiceDao;
 
     @Inject
     public ExperimentController(ScxaExperimentTrader experimentTrader, DataFileHub dataFileHub,
                                 ExperimentPageContentService experimentPageContentService,
-                                ExperimentAttributesService experimentAttributesService) {
+                                ExperimentAttributesService experimentAttributesService,
+                                TSnePlotServiceDao tSnePlotServiceDao) {
         this.experimentTrader = experimentTrader;
         this.dataFileHub = dataFileHub;
         this.experimentPageContentService = experimentPageContentService;
         this.experimentAttributesService = experimentAttributesService;
+        this.tSnePlotServiceDao = tSnePlotServiceDao;
     }
 
     @RequestMapping(value = {"/experiments/{experimentAccession}", "/experiments/{experimentAccession}/**"},
@@ -44,8 +48,8 @@ public class ExperimentController extends HtmlExceptionHandlingController {
         Experiment<Cell> experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
 
         model.addAllAttributes(experimentAttributesService.getAttributes(experiment));
-
         model.addAttribute("content", GSON.toJson(experimentPageContentForExperiment(experiment, accessKey)));
+        model.addAttribute("numberOfCells", tSnePlotServiceDao.fetchCellNumberByExperimentAccession(experimentAccession));
 
         return "experiment-page";
     }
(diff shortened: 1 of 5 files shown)


alfonsomunozpomer (changes requested):
Everything’s good: short concise commit messages and clear intent in the code. Just add a test for `fetchCellNumberByExperimentAccession` and verify that coverage of `TSnePlotServiceDao.java` is at least as high as before.

sc/src/main/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDao.java
```diff
@@ -89,4 +89,15 @@ public TSnePlotServiceDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate)
                 Integer.class
         );
     }
+
+    private static final String COUNT_CELLS_BY_EXPERIMENT_ACCESSION =
+            "SELECT COUNT(*) FROM scxa_tsne WHERE experiment_accession=:experiment_accession";
+    public Integer fetchCellNumberByExperimentAccession(String experimentAccession){
```
alfonsomunozpomer: Space before curly brace. Please [install CheckStyle plug-in](https://maven.apache.org/plugins/maven-checkstyle-plugin/) in IntelliJ or run it as a Maven task before committing to catch style mistakes.

sc/src/test/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDaoIT.java
```diff
@@ -103,6 +103,13 @@ void testPerplexities(String experimentAccession) {
                 .doesNotHaveDuplicates();
     }
 
+    @ParameterizedTest
+    @MethodSource("randomExperimentAccessionProvider")
+    void testNumberOfCellsByExperimentAccession(String experimentAccession) {
+        assertThat(subject.fetchCellNumberByExperimentAccession(experimentAccession))
+                .isInstanceOfAny(Integer.class);
+    }
+
```
alfonsomunozpomer, line 133: Here checking if the value is an `Integer` is ok but it could be better. You could, for instance, [assert that the value is positive](http://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractIntegerAssert.html#isPositive--).

In any case, you should test that the actual value of cells is the right one. The most straightforward way to do so would be to use SQL, but notice that this wouldn’t be of much value since you’d be repeating the same solution of the tested subject, `TSnePlotServiceDao`. You should count the number of lines in one of the t-SNE plot files. You can use the class `DataFileHub` to get any of them.

alfonsomunozpomer (changes requested):
Kudos for randomizing the experiment to test with a random provider! However the test should check that the number is exact, otherwise bugs could slip through.

alfonsomunozpomer (changes requested):
Notice that because you’re populating the t-SNE table with the new full fixture, other tests in `TSnePlotServiceDaoIT` are failing. You need to restrucutre the test class to address this.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentController.java
```diff
@@ -44,8 +48,8 @@ public String showExperimentPage(Model model,
         Experiment<Cell> experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
 
         model.addAllAttributes(experimentAttributesService.getAttributes(experiment));
-
         model.addAttribute("content", GSON.toJson(experimentPageContentForExperiment(experiment, accessKey)));
+        model.addAttribute("numberOfCells", tSnePlotServiceDao.fetchCellNumberByExperimentAccession(experimentAccession));
```
monicajianu: You could move this line inside the `experimentPageContentForExperiment` method, as the number of cells are also content.

sc/src/main/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDao.java
```diff
@@ -89,4 +89,15 @@ public TSnePlotServiceDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate)
                 Integer.class
         );
     }
+
+    private static final String COUNT_CELLS_BY_EXPERIMENT_ACCESSION =
+            "SELECT COUNT(DISTINCT(cell_id)) FROM scxa_tsne WHERE experiment_accession=:experiment_accession";
+    public Integer fetchCellNumberByExperimentAccession(String experimentAccession) {
```
monicajianu: I would change this method name to `fetchNumberOfCellsByExperimentAccession`. `CellNumber` sounds like you are referring to an ID of sorts, rather than the count.

sc/src/test/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDaoIT.java
```diff
@@ -103,6 +111,24 @@ void testPerplexities(String experimentAccession) {
                 .doesNotHaveDuplicates();
     }
 
+    @ParameterizedTest
+    @MethodSource("randomExperimentAccessionProvider")
+    void testNumberOfCellsByExperimentAccession(String experimentAccession) {
+        cleanDatabaseTables();
```
monicajianu, line 119: You don't need to call `cleanDatabaseTables` and `populateDatabaseTables` yourself. That what the `@AfterAll` and `@BeforeAll` annotations on those methods are for. (see [here](https://junit.org/junit5/docs/current/user-guide/#writing-tests-annotations) for a description of all the possible JUnit 5 annotations)

monicajianu (commented):
A few minor things

sc/src/test/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDaoIT.java
lingyun1010: We need to call `cleanDatabaseTables` and `populateDatabaseTables` in this case, since we are using different fixure.sql dataset. We need to mount a new dataset and clear it before running the following tests. `@AfterAll` and `@BeforeAll` annotations will only run once before and after all the tests, which may not help in this case.

monicajianu: Is there any reason you can't use the same fixtures for all tests? Or you could change `@AfterAll` and `@BeforeAll` to `@AfterEach` and `@BeforeEach`.

lingyun1010: In this test, we test the count of cells. To make a comprehensive test we count the lines of the local file to match the return result by querying in the fixture. If the fixture is a partition of the full dataset, then it will fail.
Instead of changing to @AfterEach and @AfterEach, we could keep the testing speed as much as we can, since we only need to reload the dataset once.

sc/src/main/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDao.java
alfonsomunozpomer: In these cases I usually name things `blahCount` (e.g. `fetchCellCountByExperimentAccession`).

sc/src/main/webapp/WEB-INF/jsp/tiles/views/experiment-page/experiment-description.jsp
```diff
@@ -32,7 +33,7 @@
             </h3>
 
             <h5>${type}</h5>
-
+            <div>Number of cells: ${numberOfCells}</div>
```
alfonsomunozpomer: Use `<fmt:formatNumber value="${numberOfCells}"/>`, you’ll need to add `<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>` at the top.

alfonsomunozpomer (changes requested):
A minor detail about number formatting and this is good to go.

sc/src/test/java/uk/ac/ebi/atlas/tsne/TSnePlotServiceDaoIT.java
alfonsomunozpomer: @lingyun1010 Could you add a short comment at the top of this method explaining why you need a particular fixture for this test?

PR #192, opened 2018-11-02
Bugfix #161676946 – SC gene search endpoint is picky about reuest parameters order

alfonsomunozpomer:
I caught this one while testing the gene search SPA, and because `query-string` sorts parameters alphabetically (so that I ended up with `species` before `symbol`). An obscure bug!


monicajianu:
I don't understand... what is the issue here?

alfonsomunozpomer:
Queries like `http://localhost:8080/gxa/sc/json/search?foo=bar&symbol=aspm` would fail because the first request parameter key, `foo`, didn’t match any of the ID property names, and when `symbol` was going to be `anyMatch`ed the `validQueryFields` stream had been exhausted and Java complained about it.

I noticed because `queryString` was serializing the query `{symbol: 'aspm', species: 'homo sapiens'}` to `?species=homo+sapiens&symbol=aspm`.

monicajianu (approved):
Obscure indeed!

PR #197, opened 2018-11-07
Feature #159265221,#159265231 – Gene search SPA

alfonsomunozpomer:
Minimal SPA with a single route that uses the beta version of the faceted search results package in order to display the matching gene ID (via `resultsMessageFormatter` callback).


lingyun1010 (commented):
I have tested on `Pten` marker gene search, which displays `undefined` in the result description but the json payload is `matchingGeneId: ENSMUSG00000013663`.

alfonsomunozpomer:
@lingyun1010 I forgot to compile the bundles with the latest changes. Could you test again, pleae?

lingyun1010 (approved):
Works well now.

PR #199, opened 2018-11-08
Feature #161749807 - Add support for link in the CardModel description

monicajianu:
- Changed CardModel description from a string to a pair containing text and an optional URL. 
- Changed the adapter class to convert the pair into a JSON object.
- Changed related tests.

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
@@ -14,8 +14,10 @@ public class CardModelAdapter {
         result.addProperty("iconType", cardModel.iconType().name().toLowerCase());
         result.addProperty("iconSrc", cardModel.iconSrc());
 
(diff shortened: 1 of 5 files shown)


base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
```diff
@@ -36,15 +38,15 @@ private static JsonArray getContent(List<Pair<String, Optional<String>>> textAnd
         JsonArray content = new JsonArray();
 
         textAndUrls.forEach(textAndUrl -> content.add(
```
alfonsomunozpomer, line 40: This is one of the reasons why I’m leaning lately to creating collections (lists or maps) and then having `Gson` deal with them in the controller. It’s a bit of a shortcoming on the part of Gson’s API since it forces you to populate JSON elements using side-effects.

alfonsomunozpomer (approved):
All good!

PR #200, opened 2018-11-12
Bugfix #161845714 - Marker gene links not always correct

monicajianu:
- Fixed issue where ks and cluster IDs were retrieved for all the genes in any experiment that happened to have the minimum p-value
- Fixed issue where marker genes weren't retrieved at all for experiments that don't have a preferred K
- Incresed marker gene threshold back up to 0.05
- Fixed minor code style issues and improved method naming even further

At the moment, this code will retrieve all the ks and cluster IDs where that gene has the minimum p-value. One way to mitigate this is to only retrieve the cluster ID for the preferred K. The issue is what happens when we don't have a preferred K... Do we retrieve the first K or do we show all of them?

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
@@ -28,7 +28,7 @@ import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollecti
 @Component
 public class GeneSearchDao {
     private static final Logger LOGGER = LoggerFactory.getLogger(GeneSearchDao.class);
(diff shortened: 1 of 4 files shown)


lingyun1010 (approved):

lingyun1010 (commented):
I got 17 tests failed in this branch. Could you please have a look into those?

monicajianu:
@lingyun1010 Everything is passing in Bamboo... Are you running the tests locally? If so, which ones?

lingyun1010:
@monicajianu 16 Tests failing due to my solr environment and another test is wrong in `JsonGeneSearchControllerWIT` and I will fix it in a new, upcoming branch.

alfonsomunozpomer (approved):

PR #201, opened 2018-11-13
Bugfix #160983044 - Baseline and differential tabs enabled when no results

monicajianu:
- Added filtering for expression levels (for baseline) and log fold change and p-value (for differential), when retrieving the experiment types to be shown in the gene search results page.
- Improved testing.


alfonsomunozpomer (approved):
Looks good. I think this one’s been lurking around unnoticed for a few releases. 😨

PR #205, opened 2018-11-21
Bugfix #162100444 – Retrieval of marker gene profiles without preferred k

alfonsomunozpomer:
This one fell through the cracks!

I found about this thanks to E-EHCA-2. The gist of it is that [Google’s immutable collections don’t allow null values](https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained) (it’s an interesting read, so I encourage you to take the time to read it). Unpacking an optional with `.orElse(null)` isn’t the best way of using optionals, because you’re missing all the goodness.

In terms of design it might be more correct having a second method that searches without a preferred K, but I think it’s an unnecessary complication (I left a comment to remind future Atlas developers why 0 is passed as a placeholder).

I also removed the `k IS NULL` from the query since there are no null values. I suspect this was some leftover from an earlier version of the query using `JOIN`.


sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -102,15 +102,15 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
             "WHERE experiment_accession=:experiment_accession " +
                 "AND gene_id=:gene_id " +
                 "AND marker_probability<:threshold " +
-                "AND ((k is NULL OR k=:preferred_K) " +
+                "AND (k=:preferred_K " +
```
monicajianu, line 105: Hmm I think that was meant to be `:preferred_k IS NULL` (see [here](https://stackoverflow.com/questions/21062148/how-to-query-postgres-on-optional-params))

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
```diff
@@ -84,7 +85,8 @@ public GeneSearchService(GeneSearchDao geneSearchDao, TsnePlotSettingsService ts
                     geneSearchDao.fetchClusterIdsWithPreferredKAndMinPForExperimentAccession(
                             geneId,
                             experimentAccession,
-                            preferredK.orElse(null));
+                            // If there’s no preferred k we use 0, which won’t match any row
+                            preferredK.orElse(0));
```
monicajianu, line 89: I think neither `null` nor `0` are elegant approaches here. I think separate queries would make things a bit clearer and more explicit... We could improve upon this at a later date of course.

monicajianu (approved):
Good catch! I should've done more thorough testing when I fixed the other issue 😟 . I knew E-EHCA-2 was special...

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
alfonsomunozpomer: I’m not entirely sure about that... We want this query to return the following:
1. If we pass a valid `preferredK`, two rows: one with a matching *k* and the cluster ID and another with the *k* and cluster ID with the lowest *p*-value (they could be the same).
2. If we pass `null` (or 0): only the *k* and cluster ID with the lowest *p*-value.

If we have `:preferred_k IS NULL`, in the second case we would get all the rows where the *p*-value is below the threshold since the third clause would evaluate to true.

Unless I’m missing something here...

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchService.java
alfonsomunozpomer: Agreed, in the interest of time I’ll add a ticket to Pivotal and leave it as it is if you don’t have anything against it. @lingyun1010?

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
monicajianu: Yeah, I think that's correct... Yeah maybe the two statements used to be AND-ed together? I don't know... Oh, something else to note is that this statement won't necessarily return just 2 results... There are cases where multiple k's have the same minimum probability, it will return all of them... Otherwise how do you choose which `k` to discard?

alfonsomunozpomer: I guess that the most probable case of that happening is if the *p*-value is 0 (otherwise we’re being astronomically unlucky, but yes, it’s possible). Which k should we pick in that case? The more I think about this, the more I lean towards making two SQL queries...

lingyun1010 (approved):

sc/src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
monicajianu: It's not just when the p-value is 0 (though it's close). Gene `ENSG00000167748` in `E-MTAB-5061` has the same (minimum) p-value for k = 7, 8, 9, 10, 11, 12, 13, 17... Unless the query is wrong.

alfonsomunozpomer: I see. It turns out that this happens when it’s the same cluster, which makes sense:
• https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-81547/Results?geneId=ENSG00000167748&k=13&clusterId=[4]
• https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-81547/Results?geneId=ENSG00000167748&k=14&clusterId=[8]

We’ll need to discuss which k is then preferred.

PR #208, opened 2018-11-22
Feature #161750271 – Add headers to gene search SPA

alfonsomunozpomer:
Companion PR of https://github.com/ebi-gene-expression-group/scxa-faceted-search-results/pull/12. This one ties everything together.


sc/src/main/javascript/bundles/gene-search/src/ExperimentTableCard.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
+import formatNumber from 'format-number'
 import EbiSpeciesIcon from 'react-ebi-species'
 
+const _formatNumber = formatNumber()
```
lingyun1010, line 7: Why do we need to define a `const` for `formatNumber()` instead of using it directly?

lingyun1010 (commented):
Not sure about a variable definition. The title itself is beautiful.

sc/src/main/javascript/bundles/gene-search/src/ExperimentTableCard.js
alfonsomunozpomer: `formatNumber()` returns a function which we then apply to each number. Otherwise we would need to do `formatNumber()(numberOfAssays)`, which incurs in an unnecessary (although negligible) overhead. I used it to be explicit about what `formatNumber` returns.

lingyun1010: What I mean by using it directly is actually `formatNumber()(NumberOfAssays)`, which I would put personally since we only use it once in this case.  But I think it is a personal style issue, so I will approve the branch with the current version.

lingyun1010 (approved):

alfonsomunozpomer:
Thank you @lingyun1010. Notice that if you use `formatNumber()(NumberOfAssays)` you’d be evaluating `formatNumber()` for every rendered row, whereas by having it outside that evaluation happens only once. Besides a matter of style it reflects also that it’s similar to e.g. a static method in Java.

PR #213, opened 2018-11-26
Feature/161508497 add panel view for homepage

lingyun1010:
This branch involves two main tasks relative to the new single-cell homepage design:
1. An endpoint controller to provide the latestExperiment JSON payload
2. Update homepage jsp with the newly bundled component


sc/src/main/java/uk/ac/ebi/atlas/home/JsonLatestExperimentController.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+import uk.ac.ebi.atlas.controllers.JsonExceptionHandlingController;
+import uk.ac.ebi.atlas.trader.ScxaExperimentTrader;
+import javax.inject.Inject;
+
+@RestController
+public class JsonLatestExperimentController extends JsonExceptionHandlingController {
+    private final LatestExperimentsService latestExperimentsService;
+    @Inject
```
alfonsomunozpomer: In Spring, if a `@Component` class has only one constructor, it applies field injection by default.

[`@RestController` is an annotation that combines `@Controller` and `@ResponseBody`](https://docs.spring.io/spring/docs/4.3.9.RELEASE/javadoc-api/org/springframework/web/bind/annotation/RestController.html). In turn, [`@Controller` is a specialized `@Component`](https://docs.spring.io/spring/docs/4.3.9.RELEASE/javadoc-api/org/springframework/stereotype/Controller.html). So all RestControllers are Spring components.  You don’t need to annotate it with `@Inject`.

sc/src/main/java/uk/ac/ebi/atlas/home/JsonLatestExperimentController.java
```diff
@@ -0,0 +1,22 @@
...
+    public JsonLatestExperimentController (LatestExperimentsDao latestExperimentsDao, ScxaExperimentTrader experimentTrader) {
```
alfonsomunozpomer: You should inject here the `latestExperimentsService`, annotate the service to be a `@Component` and have the DAO and the trader injected there, so [the IoC chain](https://docs.spring.io/spring/docs/4.3.x/spring-framework-reference/html/overview.html#overview-dependency-injection) follows naturally.

sc/src/main/java/uk/ac/ebi/atlas/home/LatestExperimentsService.java
```diff
@@ -10,6 +10,8 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
+
 public class LatestExperimentsService {
```
alfonsomunozpomer, line 17: See comment in the controller class.

sc/src/test/java/uk/ac/ebi/atlas/home/JsonLatestExperimentControllerWIT.java
```diff
@@ -0,0 +1,72 @@
+package uk.ac.ebi.atlas.home;
```
alfonsomunozpomer, line 1: Please, space lines following our conventions (see other tests). Same thing applies to `LatestExperimentDaoIT.java`.

sc/src/main/java/uk/ac/ebi/atlas/home/LatestExperimentsService.java
```diff
@@ -39,6 +41,10 @@ public LatestExperimentsService(LatestExperimentsDao latestExperimentsDao,
         this.experimentTrader = experimentTrader;
     }
 
+    public String fetchLatestExperimentsJSON(){
```
alfonsomunozpomer: Any new public methods should be tested. To test services, please mock the DAO and trader and write some unit tests.

alfonsomunozpomer (changes requested):
Spring dependency injection mechanism is a bit hard to grasp at first. Restructure the constructors so that each class can inject the necessary fields. Polish the test code and add tests for the service.

All in all, I can see a better design than the initial version.

sc/src/test/java/uk/ac/ebi/atlas/home/JsonLatestExperimentControllerWIT.java
alfonsomunozpomer: This is hardly enough.

Again, have a look at other tests and use them as a template. You should be more meticulous. Having readable, maintainable, clean code is our top priority here. You’ll be saving two minutes but adding more than that to future Atlas developers. Please, install the CheckStyle plug-in in IDEA, since that will help you with style fixes.

sc/src/main/java/uk/ac/ebi/atlas/home/JsonLatestExperimentController.java
alfonsomunozpomer: Remember to fix the constructor.

alfonsomunozpomer (changes requested):
You should read the reviews more carefully and make changings accordingly. One last push, this is almost done.

alfonsomunozpomer (changes requested):
Review the following in all your .java files:
• package line should be followed by a blank line
• static imports should be preceded and followed by a blank line above and after
• end with a blank line

Take the time to install the CheckStyle plug-in.

sc/src/main/webapp/WEB-INF/tiles/views.xml
```diff
@@ -11,7 +11,7 @@
                 <put-attribute name="home-hero" value="/WEB-INF/jsp/tiles/views/home/hero.jsp"/>
                 <put-attribute name="search" value="/WEB-INF/jsp/tiles/views/home/search.jsp"/>
                 <put-attribute name="experiment-list-latest-box"
```
alfonsomunozpomer: This section of the JSP should be renamed, since it’s not only the the latest experiments.

sc/src/main/webapp/WEB-INF/tiles/views.xml
```diff
@@ -10,8 +10,8 @@
             <definition template="/WEB-INF/jsp/tiles/views/home.jsp">
                 <put-attribute name="home-hero" value="/WEB-INF/jsp/tiles/views/home/hero.jsp"/>
                 <put-attribute name="search" value="/WEB-INF/jsp/tiles/views/home/search.jsp"/>
-                <put-attribute name="experiment-list-latest-box"
-                               value="/WEB-INF/jsp/tiles/views/home/experiment-list-latest.jsp"/>
+                <put-attribute name="experiment-and-species-panels"
```
alfonsomunozpomer, line 13: If you change the attribute name here you should change it in the corresponding view as well.

sc/src/main/webapp/WEB-INF/jsp/tiles/views/home/home-panels.jsp
```diff
@@ -0,0 +1,42 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
```
alfonsomunozpomer: JSPs inside the directory `views/home` are HTML fragments for the home page. You shouldn’t include any `<html>`, `<head>` or `<body>` tags (have a look at the other fragments in this directory). Also, remove references to CSS files that are included in the base template (`base.jsp`). Also, notice that you’re adding a style from Expression Atlas, and you should only use the styles from Single Cell Expression Atlas. If there are styles missing add them accordingly (be careful about copy & paste and add only and strictly what you need).

alfonsomunozpomer (changes requested):
The Java part is now done, but as it is, instead of the home page you get an error message. Be sure to test everything before pushing, innocent changes can have bad consequences. Clean up the HTML and I think this is done.

sc/src/main/webapp/resources/css/atlas.css
```diff
@@ -55,4 +55,305 @@ hr {
 /* Match margin set by React-Select components */
 #experiment-page .gene-search-container {
     margin-bottom: 1rem;
+}
+
```
alfonsomunozpomer, line 59: Have you tested that we actually need all theses classes in our CSS files?

sc/src/test/java/uk/ac/ebi/atlas/home/LatestExperimentsServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final ThreadLocalRandom RNG = ThreadLocalRandom.current();
+  private LatestExperimentsService subject;
+
+  @BeforeEach
+  void setUp() {
+    subject = new LatestExperimentsService(latestExperimentsDaoMock, experimentTraderMock);
+  }
+
+  @Test
+  void expectedClustersOnlyInIdf() {
```
alfonsomunozpomer: I think you forgot to rename this test! ;)

alfonsomunozpomer (approved):
I can’t spot any other issues. Feel free to squash & merge.

PR #214, opened 2018-11-30
Feature #161607744 – Suggest two char symbols

alfonsomunozpomer:
Besides the code changes, I removed the English stop words: `solr/admin/zookeeper?detail=true&path=%2Fconfigs%2Fbioentities%2Flang%2Fstopwords_en.txt`. There are many such as *the*, *in*, *as*... that return results. This is an infrastructure change that we need to propagate to other environments if we want results from these queries (it’s not critical since Laura is only interested in AR ;) ).

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterDao.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterDao.java
@@ -7,6 +7,7 @@ import uk.ac.ebi.atlas.solr.bioentities.query.BioentitiesSolrClient;
 import uk.ac.ebi.atlas.species.Species;
 
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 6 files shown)


base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterDao.java
```diff
@@ -56,7 +56,7 @@ public SuggesterDao(BioentitiesSolrClient solrClient) {
                 .setParam("suggest.dictionary", suggesterDictionary)
                 .setParam("suggest.q", query)
                 // We raise suggest.count to a high enough value to get exact matches (the default is 100)
-                .setParam("suggest.count", "500")
+                .setParam("suggest.count", "750")
```
monicajianu, line 59: I don't quite understand this. I would assume (naïvely, perhaps) that the suggester prioritises exact matches... So no matter how low this count is, you'd still get some.

base/src/test/java/uk/ac/ebi/atlas/search/suggester/SuggesterDaoIT.java
```diff
@@ -91,7 +91,7 @@ void closestMatchIsFirst() {
         String twoCharSymbol = "AR";
 
         // Suggestion::equals is established only by term and payload, weight isn’t considered
-        assertThat(subject.fetchBioentityProperties(twoCharSymbol.toLowerCase(), 10, false))
+        assertThat(subject.fetchBioentityProperties(twoCharSymbol, 10, false))
                 .isNotEmpty()
                 .startsWith(new Suggestion(twoCharSymbol, 0, "symbol"));
         assertThat(subject.fetchBioentityProperties("ar", 10, false, speciesUtils.getHuman()))
```
monicajianu, line 97: Should this be `twoCharSymbol.toLowerCase()`?

base/src/main/java/uk/ac/ebi/atlas/search/suggester/SuggesterDao.java
alfonsomunozpomer: Unfortunately, [the suggester implementation we use](https://lucene.apache.org/solr/guide/7_1/suggester.html#analyzinginfixlookupfactory) analyzes and tokenizes the input and performs matches against a Lucene index (which is very similar to a regular Solr query), so that’s not always the case.

There are [other](https://lucene.apache.org/solr/guide/7_1/suggester.html#analyzinginfixlookupfactory) [suggesters](https://lucene.apache.org/solr/guide/7_1/suggester.html#analyzinginfixlookupfactory) which include the boolean parameter `exactMatchFirst` but have no context filtering, which we need to filter suggestions by species. So, in principle, we could have one suggester per species, which I considered and maybe it’s better in the long term for the sake of correctness and performance. I think that for the time being this was the path of least resistance.

The 750 value was a bit of trial error, by the way...

base/src/test/java/uk/ac/ebi/atlas/search/suggester/SuggesterDaoIT.java
alfonsomunozpomer: I thought so at first, but later I realised that the suggestions for e.g. “asp” are different than “ASP” or “Asp”.

monicajianu (approved):

PR #218, opened 2018-12-07
Bugfix/162416247 add empty array if null marker genes

lingyun1010:
To make SC facet search results table header sorting works well, I add an empty array for null markerGenes attribute. It will not affect the existed work, but make sure the sorting is enabled all the time, since I check if a header is included in the response attributes.

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -160,7 +160,7 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
                                                 markerGeneFacets.get(geneId).get(experimentAccession)));
                             } else {
                                 experimentAttributes.put(
-                                        "markerGenes", new ArrayList());
+                                        "markerGenes", ImmutableList.of());
                             }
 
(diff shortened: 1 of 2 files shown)


sc/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -158,6 +158,9 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
                                                 experimentAccession,
                                                 geneId,
                                                 markerGeneFacets.get(geneId).get(experimentAccession)));
+                            } else {
+                                experimentAttributes.put(
+                                        "markerGenes", new ArrayList());
```
alfonsomunozpomer: In general it’s better to use [`Collections.emptyList()`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyList--) or Guava’s [`ImmutableList.of()`](https://google.github.io/guava/releases/24.1-jre/api/docs/com/google/common/collect/ImmutableList.html#of--) to make the emtpy list immutable and communicate your intent clearly, since we know that list can’t be modified.

alfonsomunozpomer (changes requested):
Could you add a test that verifies the new behaviour?

PR #221, opened 2018-12-20
Feature/162696999 add new smiley feedback form

lingyun1010:
This branch bundles the [smiley feedback package](https://github.com/ebi-gene-expression-group/atlas-smileys-feedback-form) and display the feedback button in both gxa/sc results and experiment pages. New feedback popup window stylesheet has been added to gxa/sc resources.


gxa/src/main/webapp/WEB-INF/jsp/tiles/local-masthead.jsp
```diff
@@ -33,7 +33,7 @@
                     <li id="local-nav-help"><a href="${pageContext.request.contextPath}/help/index.html"><i class="icon icon-generic padding-right-medium" data-icon="?"></i>Help</a></li>
                     <li id="local-nav-licence"><a href="${pageContext.request.contextPath}/licence.html"><i class="icon icon-generic padding-right-medium" data-icon="'"></i>Licence</a></li>
                     <li id="local-nav-about"><a href="${pageContext.request.contextPath}/about.html"><i class="icon icon-generic padding-right-medium" data-icon="i"></i>About</a></li>
-                    <li id="local-nav-feedback"><a href="https://www.ebi.ac.uk/support/gxa" target="_blank" data-icon="X"><i class="icon icon-functional padding-right-medium" data-icon="n"></i>Feedback</a></li>
+                    <li id="local-nav-feedback"><a href="https://www.ebi.ac.uk/support/gxa" target="_blank" data-icon="X"><i class="icon icon-functional padding-right-medium" data-icon="n"></i>Support</a></li>
```
monicajianu, line 36: Don't forget to change this in single cell as well!

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/experiment-page.jsp
```diff
@@ -1,24 +1,41 @@
 <%@ page contentType="text/html;charset=UTF-8" %>
+<%@ include file="experiment-page/experiment-description.jsp" %>
 
 <link rel="stylesheet" href="${pageContext.request.contextPath}/resources/css/experiment.css">
 <link rel="stylesheet" href="${pageContext.request.contextPath}/resources/css/alt-customized-bootstrap-3.3.5.css">
+<link rel="stylesheet" href="${pageContext.request.contextPath}/resources/css/popup-feedback.css" type="text/css" media="all">
 
-<%@ include file="experiment-page/experiment-description.jsp" %>
+<style>
+    #experiment-page, #feedback-button {
```
monicajianu: There is some repetition when adding these rules to the experiment page and the results page. Consider putting the rules in the CSS file instead.

gxa/src/main/webapp/resources/css/popup-feedback.css
```diff
@@ -0,0 +1,200 @@
+.mm-popup {
```
monicajianu, line 1: Are these CSS rules taken from somewhere? If yes, you should make sure the software license allows copying, and attribute the source appropriately. Also, given that these rules are styling the Popup component, I feel making the Popup a styled component would be more suitable. I realise that would be a massive pain, given how many rules there are.

sc/src/main/javascript/bundles/feedback-form/html/index.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
```
monicajianu: Seems like you're duplicating this from the component repository itself. I don't think it's necessary, given it's such a simple component. You can remove the `html` folder, along with the webpack and babel settings. You can take inspiration from the `autocomplete` bundle, as it's equally simple. Same comments apply for gxa.

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/experiment-page.jsp
monicajianu: Same comment applies for SC.

gxa/src/main/javascript/package.json
```diff
@@ -40,6 +40,7 @@
     "react-localstorage": "^1.0.0",
     "react-refetch": "^2.0.1",
     "react-svg": "^7.2.2",
+    "react-tooltip": "^3.9.0",
```
monicajianu, line 43: Is this needed for something?

monicajianu (changes requested):
Minor things. It would be great if we could include the popup CSS in the component!

gxa/src/main/webapp/resources/css/popup-feedback.css
lingyun1010: This is [CSS styling documentation](https://github.com/minutemailer/react-popup/blob/gh-pages/popup.example.css) is recommended to use in order for the popup to display correctly in the [react-popup package](http://minutemailer.github.io/react-popup/). I think they should be written in a separate css file rather than in js.

monicajianu: If you copy-paste someone else's code (even if it is just CSS), it is good practice to attribute the source. If nothing else, it will serve as guidance for future developers that might wonder where the file is from. 

Also, you should add a note in the [atlas-smileys-feedback-form](https://github.com/ebi-gene-expression-group/atlas-smileys-feedback-form) Readme file to explain how to style the component, and link to the CSS example. Otherwise, someone running the component might think it's broken if they don't have the Atlas CSS.

monicajianu (approved):
Looks good!

sc/src/main/webapp/WEB-INF/jsp/tiles/views/experiment-page.jsp
```diff
@@ -1,11 +1,17 @@
 <%@ page contentType="text/html;charset=UTF-8" %>
 <%@ include file="experiment-page/experiment-description.jsp" %>
 
+<link rel="stylesheet" href="http://localhost:8080/gxa/sc/resources/css/popup-feedback.css" type="text/css" media="all">
```
monicajianu: You cannot link to `localhost` here, otherwise it will not work anywhere else. You should add this CSS file the way we normally link resources, in `base.jsp`.

sc/src/main/webapp/WEB-INF/jsp/tiles/views/experiment-page.jsp
```diff
@@ -1,11 +1,17 @@
...
+<link rel="stylesheet" href="http://ebi.emblstatic.net/web_guidelines/EBI-Icon-fonts/v1.3/fonts.css" type="text/css" media="all">
```
monicajianu: Same comment as before, you must add CSS resources in the `base.jsp`, otherwise we will lose track of which CSS we use there. Saying that, we have already added the EBI icons v1.2. You cannot add v1.3 as well as they do not play nicely with each other. So try to pick an icon that is available in 1.2.

We do have another story to migrate to 1.3 (https://www.pivotaltracker.com/story/show/163226652), but it's not as straighforward as it seems, as 1.3 has a lot of breaking changes. Therefore, a migration should be done as part of a separate story.

sc/src/main/webapp/WEB-INF/jsp/tiles/views/gene-search-results.jsp
```diff
@@ -5,17 +5,34 @@
 
 <%@ page contentType="text/html;charset=UTF-8" %>
 
+<link rel="stylesheet" href="http://localhost:8080/gxa/sc/resources/css/popup-feedback.css" type="text/css" media="all">
```
monicajianu: Same comments as for the `experiment-page.jsp`. You do not need to add the CSS to every JSP, as they all inherit from `base.jsp`.

monicajianu (changes requested):
Sorry, I have had to withdraw my approval. I missed a few things in the SC part of the code when I reviewed it the first time.

monicajianu (approved):

PR #222, opened 2019-01-09
Feature/162396232 redesign ea homepage

lingyun1010:
Define a properties file that can be read/written by data production (`species/atlas-information.json`).
Create an endpoint and a new card type/factory for gxa popular species.
Use [homepage cards container](https://github.com/ebi-gene-expression-group/atlas-homepage-cards/pull/9) and [homapge panel contrainer](https://github.com/ebi-gene-expression-group/scxa-home-landing-page/pull/1).
Make featured experiments a single row with 6 cards.
Explicit baseline/differential experiments, don’t use B/D circles.


monicajianu:
@lingyun1010 Where can I find a sample `species/atlas-information.json` file? I can't run the application without it.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardIconType.java
```diff
@@ -1,5 +1,5 @@
 package uk.ac.ebi.atlas.model.card;
 
 public enum CardIconType {
-    SPECIES, IMAGE
+    SPECIES, IMAGE, ATLASSPECIES
```
monicajianu: How is `ATLASSPECIES` different to the `SPECIES` card type?

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
```diff
@@ -32,6 +33,24 @@ public CardModel create(PopularSpeciesInfo popularSpeciesInfo) {
                                 Optional.empty())));
     }
 
+
+    public CardModel createAtlasHomePageSpeciesCard(PopularSpeciesInfo popularSpeciesInfo) {
+        List content = new ArrayList();
+        content.add(Pair.of(popularSpeciesInfo.totalExperiments() + " experiments",
+                Optional.empty()));
+        content.add(Pair.of(popularSpeciesInfo.baselineExperiments() + " ",
```
monicajianu: You should add the words "baseline" and "differential" after the number of experiments, rather than just an empty space. Seeing as the B and D icons have been removed, the numbers on their own don't make much sense!

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
```diff
@@ -32,6 +33,24 @@ public CardModel create(PopularSpeciesInfo popularSpeciesInfo) {
                                 Optional.empty())));
     }
 
+
+    public CardModel createAtlasHomePageSpeciesCard(PopularSpeciesInfo popularSpeciesInfo) {
+        List content = new ArrayList();
```
monicajianu: In Java you specify the type when you instantiate a new list. So it should look something like `List<Pair<String, Optional<String>>> content = new ArrayList<>()`. Or you could use `Arrays.asList()` rather than repeatedly adding to the `content` variable, see here: https://www.geeksforgeeks.org/arrays-aslist-method-in-java-with-examples/

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
```diff
@@ -32,6 +33,24 @@ public CardModel create(PopularSpeciesInfo popularSpeciesInfo) {
                                 Optional.empty())));
     }
 
+
+    public CardModel createAtlasHomePageSpeciesCard(PopularSpeciesInfo popularSpeciesInfo) {
```
monicajianu: All in all, assuming `ATLASSPECIES` is no different to `SPECIES`, this method is very similar to the `create()` method above. Is there any way you could refactor to minimise code duplication?

base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
```diff
@@ -0,0 +1,62 @@
...
+            while (reader.hasNext()) {
+                atlasInformation.addAll(readAtlasInformation(reader));
+            }
+            reader.endArray();
+        }
+
+        return atlasInformation;
+    }
+
+    private ArrayList<String> readAtlasInformation(JsonReader reader) throws IOException {
```
monicajianu: The structure of the `atlas-information.json` needs to be more well-defined. The keys should describe the value stored in the field, rather than where the information is displayed on the page. What would happen if we provided this file for download, and a user wanted to find out the Ensembl Genomes version was in a programmatic way? They would have to parse the string stored in the `right` field and they would have to make a lot of assumptions.

Also, you should retrieve the number of species, assays and studies by querying the database. This information will not be updated by data production, as we have all the information we require.

Additionally, when you define a new schema for a file that will have to be used/updated by another team, you should document the schema somewhere on Confluence. That way everyone is clear about the expected file structure.

base/src/test/java/uk/ac/ebi/atlas/model/card/CardModelFactoryTest.java
```diff
@@ -67,6 +70,26 @@ void createPopularSpeciesCard() {
 
     }
 
+    @Test
+    void createAtlasPopularSpeciesCard() {
```
monicajianu: This test is almost exactly the same as the one above. We should aim to minimise code duplication.

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/browse-by.jsp
```diff
@@ -1,73 +1,22 @@
...
-
-    <div class="row align-row-to-bottom">
-        <div class="small-6 small-centered columns margin-top-large">
-            <a id="viewAllLink" href="${pageContext.request.contextPath}/experiments" class="button float-center">View all species</a>
-        </div>
-    </div>
-</div>
-
-<script defer src="${pageContext.request.contextPath}/resources/js-bundles/expressionAtlasBrowseBySpecies.bundle.js"></script>
+<link rel="stylesheet" href="https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/css/ebi-global.css" type="text/css" media="all" />
```
monicajianu: These stylesheets have already been added in the `base.jsp` file, so they shouldn't be added here as well.

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/experiment-list-latest.jsp
```diff
@@ -13,58 +12,56 @@
         <%--Using line height hack to align contents vertically, waiting for Flexible Grid support to align stuff inside rows: https://github.com/ebiwd/EBI-Framework/issues/71--%>
         <%--Links inside links to highlight text on hover, not necessary but aesthetically pleasing--%>
         <div class="tabs-panel is-active" id="featured-experiments">
-            <div class="row margin-bottom-xxlarge">
-                <div class="small-4 columns combo text-center padding-top-medium">
+            <div class="row margin-bottom-xxlarge" style="max-width: none;">
+                <div class="small-2 columns combo text-center padding-top-medium">
                     <a href="${pageContext.request.contextPath}/experiments?experimentSet=ENCODE">
-                        <img class="margin-bottom-large experiment-thumbnail" src="${pageContext.request.contextPath}/resources/images/experiment-list-latest/encode.png" />
+                        <img class="margin-bottom-large experiment-thumbnail" style="height: 100px;" src="${pageContext.request.contextPath}/resources/images/experiment-list-latest/encode.png" />
```
monicajianu: The featured experiment thumbnail are really too large, they should be a bit smaller. As it is they are taking over the page. Also, when you hover over some of them you can see a thick black line (see screenshot).
screenshot 2019-01-17 at 10 52 35

It would also be nice if the thumbnails and the container were responsive. When you shrink the page, some of them appear rather squished:
screenshot 2019-01-17 at 10 55 42

Maybe 2 rows of 6 elements is not the best approach if the screen size is smaller?

sc/src/main/javascript/bundles/scxa-homepage-panels/html/index.html
```diff
@@ -24,15 +24,18 @@
 <!-- Set to http://localhost:8080/gxa/ or http://localhost:8080/gxa_sc/ -- Remember the trailing slash! -->
```
monicajianu: This should not be part of this PR. Please remove it and all SC-related code and make a separate PR for changes related to Single Cell.

monicajianu (changes requested):
Good start, a few changes are required though.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardIconType.java
lingyun1010: atlas species card has differential and baseline button

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
lingyun1010: I think you have misunderstood this branch. The design of `ATLASSPECIES` card is [homepage cards container](https://github.com/ebi-gene-expression-group/atlas-homepage-cards/pull/9), where you can see the number is behind `Baseline` and `Differential` icons. This atlas branch is just used to pass the json file for cards container for now, but have not bundled the unreviewed [homepage cards container](https://github.com/ebi-gene-expression-group/atlas-homepage-cards/pull/9).

base/src/main/java/uk/ac/ebi/atlas/model/card/CardIconType.java
monicajianu: This enum does not denote the card type, but the type of icon used in the card. I.e. a `SPECIES` icon type is represented by the EBI species icons, and a `IMAGE` icon type will make use of static image assets such as `.png` files. How is the `ATLASSPECIES` icon type different?

lingyun1010: Because we need to render different type of cards when we received the card type from backend. `ATLASSPECIES` specified that it needs a species card with B and D links, which is different from the rendering of `SPECIES` cards.

base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
lingyun1010: I have changed the json file as following.
```
[{
  "Ensembl": 94, 
  "Ensembl Genomes": 41, 
  "WormBase paraSite": 11, 
  "EFO": 2.98
}]
```

monicajianu: Much better! However, while there isn't an established standard for naming JSON keys, most conventions dictate that you shouldn't use spaces in the keys. A lot of parsers can't really handle them. Remember also that this file will be updated programmatically, as part of a pipeline. We want to make it as easy to parse as possible.

Have a look at other JSON objects we produce, either through endpoints or in files (for example, `species-properties.json`). Can you spot a naming convention for the keys?

Also, your object is wrapped in an array. Is that necessary? Do you envision other objects being stored in this file?

lingyun1010: Yes. Thank you for your great comments. I have changed the json file into 
```
{
  "ensembl": 94, 
  "genomes": 41, 
  "paraSite": 11, 
  "efo": 2.98
}
```

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/experiment-list-latest.jsp
lingyun1010: Nice point. I have changed it to 3 elements when the screen is small.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardIconType.java
lingyun1010: I have made corresponding changes to use the `SPECIES` card type only in gxa and sc homepages.
Atlas homepage screenshot

gxa/src/main/java/uk/ac/ebi/atlas/home/HomeController.java
```diff
@@ -76,17 +103,27 @@ public String getHome(Model model) {
...
-        model.addAttribute(
-                "animalsList", GSON.toJson(popularSpeciesService.getPopularSpecies("animals", FEATURED_SPECIES)));
-        model.addAttribute(
-                "plantsList", GSON.toJson(popularSpeciesService.getPopularSpecies("plants", FEATURED_SPECIES)));
-        model.addAttribute(
-                "fungiList", GSON.toJson(popularSpeciesService.getPopularSpecies("fungi", FEATURED_SPECIES)));
-
         model.addAttribute("speciesPath", ""); // Required by Spring form tag
 
+        JsonArray experimentsjsonArray = experimentInfoListService.getExperimentsJson().get("aaData").getAsJsonArray();
```
monicajianu: I think you could get data straight from services and DAO classes, as opposed to parsing the JSON in the backend. If you think about it, you're going from Java objects -> JSON -> Java objects. It would make your life much easier if you skipped the JSON parsing in the middle! In order to do this, you could make use of the  `listPublicExperiments()` method in the `ExperimentInfoListService`, for example.

base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public ArrayList<String> fetchAll() throws IOException {
+        ArrayList<String> atlasInformation = new ArrayList<>();
+        try (JsonReader reader = atlasInformationJsonFile.get()) {
+            reader.beginObject();
+            while (reader.hasNext()) {
+                String name = reader.nextName();
+                switch(name){
+                    case "ensembl": atlasInformation.add("Ensembl: " + reader.nextString()); break;
```
monicajianu: It is not the DAO's responsibility to format the data, its only responsibility is to retrieve it. The UI is responsible for saying what this data means. The side-effect of this implementation is that you end up doing strange things in the `.jsp` to add commas. What would you do if you wanted to display the EFO version, for example, on a different page altogether? You'd have to parse each string in the list and find out which one is the EFO version, then extract the number.

Can you think of a better data structure that would allow you to retrieve the value for any given piece of metadata? Also, can you think of a way to identify the data without having to pass strings around? Hint: read about [enums in Java](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).

base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Inject
+    public AtlasInformationDao(Path atlasInformationFilePath) {
+        atlasInformationJsonFile =
+                new JsonFile.ReadOnly(
+                        atlasInformationFilePath.getParent().toString(),
+                        atlasInformationFilePath.getFileName().toString());
+    }
+
+    public ArrayList<String> fetchAll() throws IOException {
```
monicajianu: I know you modelled this on `SpeciesPropertiesDao`, but I think the JSON parsing logic could be simplified a LOT if you used [Gson](https://google.github.io/gson/apidocs/com/google/gson/Gson.html#fromJson-java.io.Reader-java.lang.Class-). So your code could be as simple as:
```
 JsonReader reader = new JsonReader(new FileReader(atlasInformationFilePath.toString()));
 SomeClass data = GsonProvider.GSON.fromJson(reader, SomeCLass.class);
```

@alfonsomunozpomer Is there are a reason why the `SpeciesPropertiesDao` is so convoluted?

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
```diff
@@ -21,15 +22,29 @@ public CardModelFactory(UrlHelpers urlHelpers) {
...
+    public CardModel create(PopularSpeciesInfo popularSpeciesInfo) {
+        List<Pair<String, Optional<String>>> content = new ArrayList<>();
+        content.add(Pair.of(popularSpeciesInfo.totalExperiments() + " experiments",
+                Optional.empty()));
+        if(popularSpeciesInfo.baselineExperiments() > 0){
+            content.add(Pair.of("Baseline: " + popularSpeciesInfo.baselineExperiments(),
+                Optional.of(urlHelpers.getExperimentsFilteredBySpeciesAndExperimentType(popularSpeciesInfo.species(), "baseline"))));
+        }
+        if(popularSpeciesInfo.differentialExperiments() > 0){
+            content.add(Pair.of("Differential: " + popularSpeciesInfo.differentialExperiments() + " ",
```
monicajianu: You don't need the blank space you append at the end of this string.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelFactory.java
```diff
@@ -21,15 +22,29 @@ public CardModelFactory(UrlHelpers urlHelpers) {
...
+                Optional.of(urlHelpers.getExperimentsFilteredBySpeciesAndExperimentType(popularSpeciesInfo.species(), "differential"))));
+        }
+
+        return createCardModel(popularSpeciesInfo, CardIconType.SPECIES, content);
```
monicajianu: You only call this private `createCardModel` method once. Was the intention to call it from `createLandingPageImageCard` as well?

monicajianu (changes requested):
I added a few more comments. Don't forget to run Checkstyle on the code to spot formatting issues. Also, keep an eye on Bamboo! There are a few failing tests for this branch.

base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
lingyun1010: Thank you for your comments. I have changed the data structure into Hashmap to retrieve any piece of metadata. However, I did not see the point of `enums` usage here, since the data we read from the file is the type of `String`, why we do not just pass strings around directly?

monicajianu: What happens if we change the structure of the file? Let's say we want to rename "paraSite" to "Fred". Let's also imagine that you used the string "paraSite" to extract the information from the map in 3 different classes. You'll end up having to change 3 different files to change the property name. However, if you use enums, you only need to change 1 file, i.e. the definition of the enum. An enum would also allow you to make use of more advanced features of your IDE to find out where they're used. If you use strings, your only option is using the Find function to search across all the files in the project. 

You're probably thinking, "well I know exactly which files need changing, I won't need to use Find anyway". But what if 6 months have gone by? What if it's not you changing the files, but another developer?

lingyun1010: Wow, I see. I like your "probably thinking" paragraph. :) I will have a corresponding change. Thank you for your detailed description.

gxa/src/main/java/uk/ac/ebi/atlas/home/HomeController.java
```diff
@@ -76,17 +104,34 @@ public String getHome(Model model) {
...
-        model.addAttribute(
-                "plantsList", GSON.toJson(popularSpeciesService.getPopularSpecies("plants", FEATURED_SPECIES)));
-        model.addAttribute(
-                "fungiList", GSON.toJson(popularSpeciesService.getPopularSpecies("fungi", FEATURED_SPECIES)));
-
         model.addAttribute("speciesPath", ""); // Required by Spring form tag
 
+        List<ExperimentInfo> experimentlist = experimentInfoListService.listPublicExperiments();
+
+        Double numberOfAssays = 0.0;
```
monicajianu: Why is `numberOfAssays` a `Double`, which you then round later on?

lingyun1010: The data I retrieved from experimentInfoListSevice is `Double`. I want to get an integer then.

gxa/src/main/java/uk/ac/ebi/atlas/home/HomeController.java
```diff
@@ -76,17 +104,34 @@ public String getHome(Model model) {
...
+        model.addAttribute("numberOfAssays", Math.round(numberOfAssays));
+        model.addAttribute("numberOfSpecies", numberOfSpecies.size());
+
+        Map<String, String> atlasInformation = atlasInformationDao.fetchAll();
+        model.addAttribute("info", atlasInformation.toString()
+                .replace("ensemble", "Ensembl")
+                .replace("genomes", "Ensembl Genomes")
+                .replace("paraSite", "WormBase paraSite")
+                .replace("efo", "EFO")
+                .replace("=",":")
```
monicajianu: If you find yourself doing any kind of manipulation on a string, ask yourself if there are better approaches for what you are trying to do. It is not necessary to build the sentence "Ensemble x, Esembl Genomes y, etc etc" in the controller. Why can't you use the same clean approach as you did to display the number of studies and assays?

gxa/src/main/java/uk/ac/ebi/atlas/home/HomeController.java
monicajianu: I can't see any `Double`s in the `ExperimentInfo` class. 

screenshot 2019-01-31 at 13 01 51

monicajianu (approved):
Looks good! Great work! 😃

PR #225, opened 2019-02-20
Feature/163388496 implement file download endpoint

lingyun1010:


base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
```diff
@@ -388,6 +404,59 @@ public SingleCellExperimentFiles getSingleCellExperimentFiles(String experimentA
                             SINGLE_CELL_MATRIX_MARKET_TPMS_CELL_IDS_FILE_PATH_TEMPLATE,
                             experimentAccession);
 
+            rawCountsMatrix =
```
monicajianu: If I'm not mistaken, we don't raw counts for anything, so they shouldn't be added here. We don't want to make the DataFileHub bigger than it already is!

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +93,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.addHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + archiveName);
+        response.setContentType("application/zip");
+        ZipOutputStream zipOutputStream = new ZipOutputStream(response.getOutputStream());
+
+        for (String experimentAccession : array){
+            Experiment experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
+
+            List<Path> paths = new ArrayList<>();
+            paths.addAll(experimentFileLocationService.getFilePathsForArchive(experiment.getAccession(), ExperimentFileType.fromId("quantification-filtered")));
```
monicajianu: The `getFilePathsForArchive` method asks for an `ExperimentFileType` enum. You can simply it a lot: rather than doing `ExperimentFileType.fromId(...)` just write `ExperimentFileType.QUANTIFICATION_FILTERED` instead!

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +93,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
 
         zipOutputStream.close();
     }
+
+    @RequestMapping(value = "experimentlist/{experimentAccessionArray}/download/zip",
```
monicajianu: Having a list as a path parameter in the middle of a URL is a little bit unusual. I think having this as a query parameter would probably be more standard.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +93,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+            method = RequestMethod.GET,
+            produces = "application/zip")
+    public void
+    downloadMultipleExperimentsArchive(HttpServletResponse response,
+                    @PathVariable String experimentAccessionArray,
+                    @RequestParam(value = "accessKey", defaultValue = "") String accessKey) throws IOException {
+
+        String[] array = experimentAccessionArray.split("\\,");
```
monicajianu: Try to avoid any kind of string manipulation unless you really, really have to. Spring MVC is nice and powerful, take advantage of that. Have a look at this guide (especially section 7) and see if you can figure out a better type for your parameter that means you don't have to split strings: https://www.baeldung.com/spring-request-param

sc/src/test/java/uk/ac/ebi/atlas/download/ExperimentFileLocationServiceIT.java
```diff
@@ -214,4 +258,30 @@ private void existingArchiveFilesOfType(String experimentAccession,
                 .isNotEmpty()
                 .containsAnyElementsOf(fileNames);
     }
+
+    private void existingArchiveFoldersOfFilesOfType(String experimentAccession,
```
monicajianu: You cannot just copy and paste implementations and only change one line of code... This method is 99% the same as the one above! Add to the existing implementation to make it work for a new use case. We want to minimise code duplication as much as possible.

monicajianu (changes requested):
Looking good. I highlighted a few issues here and there. Focus on minimising code duplication by adapting the existing code to serve your purpose.

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
lingyun1010: This is the new file renamed from `genes_scran`, which is included in the download files.

monicajianu: If I am not mistaken, the new file is `aggregated_filtered_normalised_counts.mtx`, i.e. `SINGLE_CELL_MATRIX_MARKET_NORMOLISED_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE`. `SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE` is a different one that we don't serve as a download anywhere.

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
```diff
@@ -68,18 +68,25 @@
...
             "{0}/{0}.aggregated_counts.mtx";
-    static final String SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_GENE_IDS_FILE_PATH_TEMPLATE =
+    private static final String SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_GENE_IDS_FILE_PATH_TEMPLATE =
             SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE + "_rows";
-    static final String SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_CELL_IDS_FILE_PATH_TEMPLATE =
+    private static final String SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_CELL_IDS_FILE_PATH_TEMPLATE =
             SINGLE_CELL_MATRIX_MARKET_RAW_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE + "_cols";
 
-    static final String SINGLE_CELL_MATRIX_MARKET_FILTERED_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE =
+    private static final String SINGLE_CELL_MATRIX_MARKET_NORMOLISED_AGGREGATED_COUNTS_FILE_PATH_TEMPLATE =
```
monicajianu: Small spelling mistake, should be "normalised" not "normolised" 🙂

base/src/main/java/uk/ac/ebi/atlas/resource/DataFileHub.java
lingyun1010: Right. My mistake. I will delete them.

sc/src/test/java/uk/ac/ebi/atlas/download/ExperimentFileLocationServiceIT.java
```diff
@@ -192,7 +236,8 @@ private void existingFileOfType(String experimentAccession, ExperimentFileType f
 
     private void existingArchiveFilesOfType(String experimentAccession,
                                             ExperimentFileType fileType,
-                                            List<String> expectedFileNames) {
+                                            List<String> expectedFileNames,
+                                            Boolean folder) {
```
monicajianu: What does `folder` mean? Make sure you name your variables clearly. Also, we generally name booleans like `isSomething` or `hasSomething`.

monicajianu (approved):

monicajianu:
In principle, it looks good. I think there is still a bit of repeated logic. We now have 2 endpoints, one gives you some files in an archive given a single experiment accession, and the other essentially gives you pretty much the same, but for multiple experiment accession. I envision this could be refactored to a single endpoint, where you can either request a specific file, or all the available files, for one or many experiment accessions. Does that make sense?

Also, before you merge, maybe @alfonsomunozpomer wants to review as well?

monicajianu (changes requested):
Oops, sorry! I just noticed something. Your access key check is not quite correct... Each (private) experiment will have an access key, whereas you are expecting a single access key for potentially a list of experiments. You have two choices:

1. implement some logic to support multiple access keys (and somehow ensure they come in the same order as the experiment accessions)

2. disallow downloading of files for private experiments alltogether. I don't think this would be too much of a controversial move, given that this endpoint is intended to be used from the experiments table, which will only show public experiments. But you will have to make sure that the experiment accession you get is indeed for a public experiment, otherwise a user could download files they shouldn't have access to.

monicajianu (approved):
Thanks, approved (again)!

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +92,42 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
 
         zipOutputStream.close();
     }
+
+    @RequestMapping(value = "experimentlist/download/zip",
+            method = RequestMethod.GET,
+            produces = "application/zip")
+    public void
+    downloadMultipleExperimentsArchive(HttpServletResponse response,
+                    @RequestParam(value= "experimentAccessions", defaultValue = "") List<String> accessions,
```
alfonsomunozpomer, line 101: Because `experimentAccessions` can occur multiple times in the URL it should be `experimentAccession`. I’d even rename it to `accession`, since `experiment` is already part of the path.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +92,42 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+        String archiveName = accessions.size() + "-" + "experiments" + "-files.zip";
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.addHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + archiveName);
+        response.setContentType("application/zip");
+        ZipOutputStream zipOutputStream = new ZipOutputStream(response.getOutputStream());
+
+        for (String experimentAccession : accessions){
+            Experiment experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
+
+            List<Path> paths = new ArrayList<>();
```
alfonsomunozpomer, line 113: Use immutable collections as much as possible. Specify it also in your varaible declaration. In this case, for instance:
```java
ImmutableList<Path> paths = ImmutableList.<Path>builder()
    .addAll(experimentFileLocationService.getFilePathsForArchive(experiment.getAccession(), ExperimentFileType.QUANTIFICATION_FILTERED))
    .addAll(...)
    .add()
    .build()
```

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -122,4 +126,25 @@ void downloadArchiveForInvalidExperimentFileType() throws Exception {
...
+        this.mockMvc.perform(get(ARCHIVE_DOWNLOAD_LIST_URL)
+                .param("experimentAccessions",  EXPERIMENT_ACCESSION_LIST.toString().replace("[","").replace("]", "")))
+                .andExpect(status().isOk())
+                .andExpect(
+                        header().string(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + expectedArchiveName))
+                .andExpect(content().contentType("application/zip"));
+    }
+
+    @Test
+    void downloadArchiveForInvalidExperimentAccessionList() throws Exception {
```
alfonsomunozpomer, line 144: What happens if I enter a mixed list of valid and invalid experiment accessions? Probably not what you’d expect!

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +92,42 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
 
         zipOutputStream.close();
     }
+
+    @RequestMapping(value = "experimentlist/download/zip",
```
alfonsomunozpomer, line 96: We agreed on using hyphens for multiword URLs. Rename the path to `experiment-list/...`. Also, not really sure if `list` serves any purpose here. I’d keep it simple.

alfonsomunozpomer (commented):
Please, simplify the controller to avoid duplication and add a test case for a mixed list of valid and invalid experiments. Look into that case carefully.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: For now, I have changed to `experiments/download/zip` to keep it simple.

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
lingyun1010: Good point! I have added an `if` condition in the controller and a test for a mix of valid and invalid experiment accessions.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: The corresponding change on front-end is in the [scxa-experiment-table](https://github.com/ebi-gene-expression-group/scxa-experiment-table/pull/1).

PR #229, opened 2019-02-28
Feature #164299307 - Update evidence codes URLs

monicajianu:
Replaced evidence code URLs with new ones for microarray and RNA-seq experiments.


suhaibMo:
Thanks @monicajianu. The new evidence codes are alive.

PR #231, opened 2019-02-28
Feature #163388895 - More HCA landing page tweaks

monicajianu:
- Added Kirsten Meyer and Vlad Kiselev to list of people
- Added CZI logo to footer
- Changed people list so it's organised in two columns, rather than one big list


alfonsomunozpomer (approved):
All good. Let’s see if we can get the Elysium logo now...

PR #235, opened 2019-03-07
Feature/164437194 integrate sortable header in sc

lingyun1010:
Integrate the sortable title on sc gene search results and fix a bug in [facet-results-branch](https://github.com/ebi-gene-expression-group/scxa-faceted-search-results) to set the default sorting is based on `markerGenes` column, which is passed from top props.


alfonsomunozpomer (approved):
All good. Please remember to reindent `ExperimentCard.js` and `ExperimentsHeader.js` and merge.

PR #237, opened 2019-03-14
Feature/164403270 sync json payloads for tsnes

lingyun1010:
In this feature branch, I implement four new endpoints for gene expression json payload to synch tSne plots in SC experiment page. As the cluster tsne may be coloured by metadata or cluster, gene expression endpoints need to provide corresponding clustered data and also be compatible with an empty geneID query. In this case, four endpoints are introduced and the old endpoint for providing a single array is deleted. This new feature is visualized in the [scxa-experiment-page-tsne-plot-view](https://github.com/ebi-gene-expression-group/scxa-experiment-page-tsne-plot-view/tree/feature/164403270-sync-tsnes-with-legend).

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/experimentpage/tsne/TSnePoint.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/experimentpage/tsne/TSnePoint.java
@@ -55,6 +55,10 @@ public abstract class TSnePoint {
         return new AutoValue_TSnePoint(x, y, Optional.empty(), name, "");
     }
 
+    public static TSnePoint create(double x, double y, double expressionLevel, String name, String metadata) {
+        return new AutoValue_TSnePoint(x, y, Optional.of(expressionLevel), name, metadata);
+    }
+
     public static TSnePoint create(double x, double y, String name, String metadata) {
         return new AutoValue_TSnePoint(x, y, Optional.empty(), name, metadata);
     }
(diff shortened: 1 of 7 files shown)


monicajianu (commented):
I am lacking a little bit of context for this PR (not sure what design discussions have taken place between you and @alfonsomunozpomer), but is there any reason why an calling an endpoint like `http://localhost:8080/gxa/sc/json/experiments/E-MTAB-5061/tsneplot/20/clusters/k/3` only returns one series called "Cluster 0"...? If I am not mistaken, this will break the existing functionality for the t-SNE plot.

lingyun1010:
I did not touch the endpoint for cluster tsne plot in the backend. That example is implemented in the original code.

monicajianu:
@lingyun1010 Oops, sorry, turns out I was testing this for a `k` that we don't have in our testing data. Turns out the k's we do have for `E-MTAB-5061` are different to the ones in production!

monicajianu:
@lingyun1010  @alfonsomunozpomer I think this implementation could be simplified a lot. I feel like 4 different endpoints is overkill, seeing as the expression endpoint is now just adding the expression level and the min, max values. I don't see a reason why the two plots can't share one endpoint for clusters, and one for metadata, where you have a gene ID as an optional parameter.
Also, it is a bit odd how the expression endpoint rounds the x and y values, but the clusters endpoint doesn't. I think they should both return the same data.

alfonsomunozpomer:
This story has been superseded by the migration of the t-SNE plot to Highcharts 7 and the official React wrapper, which will use the state to keep track of displayed/hidden series in a performant way.

PR #239, opened 2019-03-18
Bugfix #164689363 – Differential species in plants landing in lowercase

alfonsomunozpomer:
I spotted  a bug buried in the plants landing page. To aggregate species (e.g. *Oryza sativa Indica group* and *Oryza sativa Japoniga group* under *Oryza sativa*) we use the reference name which is in lower case, and the `species-icon-selector.jsp` expects them to be capitalised.

Let’s merge this into the release branch and then onto `dev`.


monicajianu (approved):
Good spot! 👍

PR #240, opened 2019-03-19
Features #161089368, #162396232 – Homepage redesign

alfonsomunozpomer:
Quite a few changes, so I’ll try to summarise the main points here:
• Rename `PopularSpecies`-whatever to `SpeciesSummary`-whatever. I’ve never been too happy with that name and it serves us to, maybe in the future, get information about specific species.
• Lift the renamed `SpeciesSummary` controller logic to `atlas-core` so it can be reused by bulk and SC, changing the necessary bits by injecting a different DAO. The response is now different and the results are always grouped by kingdom. In this way the React component that displays the species is just presentational and doesn’t need to know anything about kingdom values.
• Remove the popular species fields from the model in the homepage controller.
• Add a new repo, [`species-summary-panel`](https://github.com/ebi-gene-expression-group/species-summary-panel), as a git submodule in the JS modules directory, which is bundled with [`atlas-react-fetch-loader`](https://github.com/ebi-gene-expression-group/atlas-react-fetch-loader). The bundle makes a request to the species-summary endpoint and displays the species in tabs. It also relies on the [latest version of `atlas-homepage-cards`](https://github.com/ebi-gene-expression-group/atlas-homepage-cards/pull/11).
• Follow a similar approach for latest and featured experiments, so that they are served from a single endpoint. The shame here is that, unlike species, the field `latestExperiments` can’t be serialised as cards. Maybe something we can think a bit better and imporve: if we do, we can refactor [`species-summary-panel`](https://github.com/ebi-gene-expression-group/species-summary-panel) and [`experiments-summary-panel`](https://github.com/ebi-gene-expression-group/experiments-summary-panel) to a single component.

Worthy of note is that I made the `description.text` field in the card model optional, since the featured experiments don’t have it. @monicajianu could you please change the JSON schema to reflect that?

Other improvements:
• Remove the hero image+text in SCEA.
• Use Foundation classes to lay out the release stats at the top. Remember that we shouldn’t use flex or grid and rely on the EBI VF if it’s in the environment (widgets embeddable in third parties are the exception but we have one or two of those).


base/src/main/java/uk/ac/ebi/atlas/home/SpeciesSummaryService.java
```diff
@@ -0,0 +1,60 @@
...
+    public SpeciesSummaryService(SpeciesSummaryDao speciesSummaryDao) {
+        this.speciesSummaryDao = speciesSummaryDao;
+    }
+
+    public ImmutableSortedMap<String, ImmutableList<SpeciesSummary>> getSpeciesSummaryGroupedByKingdom(int limit) {
+        return popularSpeciesInfos.get().stream()
+                        .collect(groupingBy(SpeciesSummary::getKingdom))
+                        .entrySet().stream()
+                        .collect(toImmutableSortedMap(
+                                new ArbitraryStringComparator("animals", "plants", "fungi", "protists"),
```
monicajianu: I don't think sorting of this kind is the responsiblity of the service. I think it's either the UIs responsiblity, or at the very least, the controller.

base/src/main/java/uk/ac/ebi/atlas/model/card/CardModelAdapter.java
```diff
@@ -53,4 +51,12 @@ private static JsonObject getTextAndUrl(String text, Optional<String> url) {
 
         return textAndUrl;
     }
+
+    private static JsonObject getTextAndUrl(Pair<Optional<String>, Optional<String>> description) {
```
monicajianu, line 55: Hmm there's a bit of repeated logic here when with `getLeft` and `getRight`. Could you not pass two `Optionals` to this method, the text and the url, rather than the `Pair`? That was it will also be consistent with the other implementation of the `getTextAndUrl` method.

gxa/src/main/java/uk/ac/ebi/atlas/home/JsonExperimentsSummaryController.java
```diff
@@ -0,0 +1,177 @@
...
+                ImmutableMap.of(
+                        "latestExperiments",
+                        latestExperimentsService.fetchLatestExperimentsAttributes().get("latestExperiments"),
+                        "featuredExperiments",
+                        featuredExperimentsCards().stream()
+                                .map(CardModelAdapter::serialize)
+                                .collect(toImmutableList())));
+    }
+
+    private static String getImageUrl(String imageFileName) {
```
monicajianu: These URL methods could be moved to the `UrlHelpers` class.

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/experiments-summary-panel.jsp
```diff
@@ -0,0 +1,25 @@
...
+      {
+        host: '${pageContext.request.contextPath}/',
+        resource: 'json/experiments-summary',
+        responsiveCardsRowProps: {
+          className: 'row expanded small-up-2 medium-up-3 large-up-6',
+          cardContainerClassName: 'column',
+          imageIconHeight: '3rem'
+        },
+        onComponentDidMount: function() {
+          $('#experiments-summary-panel').foundation();
```
monicajianu, line 18: 👍

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/release-stats.jsp
```diff
@@ -0,0 +1,26 @@
...
+<%--@elvariable id="numberOfAssays" type="Number"--%>
+<%--@elvariable id="genomes" type="String"--%>
+<%--@elvariable id="paraSite" type="String"--%>
+<%--@elvariable id="ensembl" type="String"--%>
+<%--@elvariable id="efo" type="String"--%>
+
+<%@ page contentType="text/html;charset=UTF-8" %>
+<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>
+
+<div class= "small-12 medium-6 columns text-left">
```
monicajianu: I think we could get away with making this text a little smaller. If you view the page on a normal-sized screen (laptop), the text takes up a second line.

Screenshot 2019-03-22 at 14 32 57

monicajianu (changes requested):
A few minor things. I'll also change the JSON schema.

gxa/src/main/webapp/WEB-INF/jsp/tiles/views/home/release-stats.jsp
alfonsomunozpomer: Yes, I noticed it too when I saw it on my laptop. I think `<h4><small>` will work. Thanks for reminding me.

alfonsomunozpomer:
@monicajianu I addressed all the issues you mentioned. I especially like how I ended up refactoring `UrlHelpers`. It forced me to rethink things a bit but the result is much better, IMO.

sc/src/main/webapp/WEB-INF/tiles/views.xml
```diff
@@ -8,6 +8,7 @@
         <put-attribute name="title" value="Home &lt; "/>
         <put-attribute name="content">
             <definition template="/WEB-INF/jsp/tiles/views/home.jsp">
+                <put-attribute name="rel" value="/WEB-INF/jsp/tiles/views/home/release-stats.jsp"/>
```
monicajianu: Can we give this a nicer, more explicit name, rather than "rel"?

gxa/src/main/webapp/WEB-INF/jsp/tiles/global-masthead.jsp
```diff
@@ -2,3 +2,10 @@
 
 <header id="masthead-black-bar" class="clearfix masthead-black-bar expanded">
 </header>
+
+<script>
+  // Hack that solves https://github.com/ebiwd/EBI-Framework/pull/139 (I got tired of waiting)
+  window.addEventListener('load', function(event) {
```
monicajianu, line 8: Hmm the side-effect of this is that there is a delay, so it starts off unexpanded and then you can see it move (or maybe my browser is particularly sluggish today). But I can't think of a better way of doing it...

monicajianu (approved):
One tiny variable name change and then it's good to go. I like the static `UrlHelpers`, I think they used to be like that when I first wrote the class 🙂

sc/src/main/webapp/WEB-INF/tiles/views.xml
alfonsomunozpomer: Oh, right! Thanks for spotting it.

gxa/src/main/webapp/WEB-INF/jsp/tiles/global-masthead.jsp
alfonsomunozpomer: I know, I’m aware of it. Once the assets are cached it’s very fast, although noticeable. It’s the kind of thing that people who are unaware of it will hardly ever notice.

Another option is to fork the framework but then we can’t use the CDN. Let’s hope the PR gets trhough, although it seemed it won’t because it falls outside their core design principles. I still think that if `row expanded` is a supported option, it should also be available on the global masthead. Otherwise pages look kind of unbalanced.

Anyway...

PR #241, opened 2019-03-21
Feature/163388496 restored implement file download endpoint

lingyun1010:


sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +91,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
         zipOutputStream.close();
     }
+
+    @RequestMapping(value = "experiments/download/zip",
+            method = RequestMethod.GET,
+            produces = "application/zip")
+    public void
+    downloadMultipleExperimentsArchive(HttpServletResponse response,
+                    @RequestParam(value= "accession", defaultValue = "") List<String> accessions,
+                    @RequestParam(value = "accessKey", defaultValue = "") String accessKey) throws IOException {
```
alfonsomunozpomer: `accessKey`should be removed in this endpoint, too. Also, if requesting multiple experiments, for which one would the key be for?

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +91,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+
+        String archiveName = accessions.size() + "-" + "experiment" + "-files.zip";
```
alfonsomunozpomer: Compact your strings as much as possible:
```java
String archiveName = accessions.size() + "-experiment-files.zip"
```

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +91,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+        response.addHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + archiveName);
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.setContentType("application/zip");
+        ZipOutputStream zipOutputStream = new ZipOutputStream(response.getOutputStream());
+
+        for (String accession : accessions){
```
alfonsomunozpomer: Be consistent with your blank spaces. Please do something to integrate CheckStyle into your workflow.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -91,4 +91,44 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+            Experiment experiment = experimentTrader.getExperiment(accession, accessKey);
+            if(experiment != null) {
```
alfonsomunozpomer: When will `experiment` be `null`?

sc/src/main/java/uk/ac/ebi/atlas/download/ExperimentFileLocationService.java
```diff
@@ -56,6 +56,18 @@ public Path getFilePath(String experimentAccession, ExperimentFileType fileType)
                         .stream()
                         .map(AtlasResource::getPath)
                         .collect(Collectors.toList());
+            case QUANTIFICATION_RAW:
+                return
+                        Arrays.asList(
```
alfonsomunozpomer: Please build lists here again with `ImmutableList.of()`. Now that you’re at it, could you change other occurrences of `Arrays.asList()`?

alfonsomunozpomer (changes requested):
Better! Some minor issues remain, please have a look.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: This is in case any experiment accession in the `RequestParam` is not valid, and the endpoint may still return the files for correct ones.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -95,19 +96,25 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                                                     @RequestParam(value = "accession", defaultValue = "") List<String> accessions) throws IOException {
 
+        List<Experiment> experiments = new ArrayList<Experiment>();
         for (String accession : accessions) {
-            Experiment experiment = experimentTrader.getExperiment(accession, "");
-            if (experiment != null) {
+            try {
+                experiments.add(experimentTrader.getPublicExperiment(accession));
+            } catch (Exception e) {
+                System.out.println("Experiment accession is not valid: " + accession);
```
alfonsomunozpomer: We don’t print anything to `System.out`. Who will see it? If you want to log any error message use a logger, which outputs messages to a file that we can inspect at any time. There are many classes that do this if you need examples.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -95,19 +96,25 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
-
-        String archiveName = accessions.size() + "-experiment-files.zip";
-        response.addHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + archiveName);
-        response.setStatus(HttpServletResponse.SC_OK);
-        response.setContentType("application/zip");
-        ZipOutputStream zipOutputStream = new ZipOutputStream(response.getOutputStream());
+    public void downloadMultipleExperimentsArchive(HttpServletResponse response,
+                                                     @RequestParam(value = "accession", defaultValue = "") List<String> accessions) throws IOException {
 
+        List<Experiment> experiments = new ArrayList<Experiment>();
```
alfonsomunozpomer: A couple of things: if you’re not going to modify the list remember to always use an `ImmutableList.Builder`.

Also, this kind of loops are better expressed using streams, variable definition and initialisation go together:
```java
ImmutableList<Experiment> experiments = accessions.stream()
    .map(experimentTrader::getPublicExperiment)
    .collect(toImmutableList())
```

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -95,19 +96,25 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
         for (String accession : accessions) {
-            Experiment experiment = experimentTrader.getExperiment(accession, "");
-            if (experiment != null) {
+            try {
+                experiments.add(experimentTrader.getPublicExperiment(accession));
+            } catch (Exception e) {
+                System.out.println("Experiment accession is not valid: " + accession);
+            }
+        }
+        if (experiments.size() >= 1) {
```
alfonsomunozpomer: A more idiomatic way of checking this condition is `!experiments.isEmpty()`.

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -38,6 +38,7 @@
 class FileDownloadControllerWIT {
     private static final String EXPERIMENT_ACCESSION = "E-MTAB-5061";
     private static final List<String> EXPERIMENT_ACCESSION_LIST = Arrays.asList("E-MTAB-5061", "E-EHCA-2");
+    private static final List<String> INVALID_EXPERIMENT_ACCESSION_LIST = Arrays.asList("E-MTAB", "E-EHCA");
```
alfonsomunozpomer: Remember to use `ImmutableList`. There is a handy shorthand constructor:
```
ImmutableList.of("E-MTAB", "E-EHCA")
```

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -133,7 +134,7 @@ void downloadArchiveForValidExperimentAccessions() throws Exception {
                 MessageFormat.format(
                         ARCHIVE_NAME, EXPERIMENT_ACCESSION_LIST.size(), "experiment");
         this.mockMvc.perform(get(ARCHIVE_DOWNLOAD_LIST_URL)
-                .param("accession", EXPERIMENT_ACCESSION_LIST.toString().replace("[","").replace("]", "")))
+                .param("accession", EXPERIMENT_ACCESSION_LIST.toString().replace("[", "").replace("]", "")))
```
alfonsomunozpomer: I can observer two problems here.

The first one is that instead of manipulating a string-serialised list you can do:
```java
String.join(",", EXPERIMENT_ACCESSION_LIST)
```
This is a lot less error prone.

The second is that we decided to change the URL format to be `?accession={accession1}&accession={accession2}` but this is setting the request parameter to `?accession={accession1},{accession2}`, isn’t it? I think that just checking HTTP status codes is insufficient in all these tests. I guess you could get the contents of the response and inspect to see that it contains what you’d expect.

alfonsomunozpomer (changes requested):
There are some comments about code conventions that we use in our codebase that we want to keep for consistency. Other tidbits are parts of the Java library API that you’re not familiar with.

The most concerning problem right now are the tests. They’re not actually checking that the downloaded zip archive is well-formed. I’m not sure if it’s possible to test the contents, but we should definitely try to. Please have a look into that.

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
lingyun1010: Yes. I see what you mean by different URL format, but Spring will parse the `?accession={accession1}&accession={accession2}` into `?accession={accession1},{accession2}`, right, so both request parameter formats are working fine.

sc/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: Yeah, but we want to use `try` and `catch` to add only valid experiments in this case, I think the code is simple by using a for loop.

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
alfonsomunozpomer: I can understand that, but you’re testing a feature that we’re not actually offering. Remember that we want to stick to a standard URL scheme. The test should build the URL by chaining `.param("accession", ...)` multiple times.

lingyun1010:
As for the content of archive I have added a test in `ExperimentFileLocationServiceIT`. In `FileDownloadControllerWIT`, the archive is not only tested by `status` but also the `filename` which is corresponding to the number of valid experiment accessions.

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -134,18 +134,21 @@ void downloadArchiveForValidExperimentAccessions() throws Exception {
...
         this.mockMvc.perform(get(ARCHIVE_DOWNLOAD_LIST_URL)
-                .param("accession", EXPERIMENT_ACCESSION_LIST.toString().replace("[", "").replace("]", "")))
+                .param("accession", EXPERIMENT_ACCESSION_LIST.get(0))
+                .param("accession", EXPERIMENT_ACCESSION_LIST.get(1)))
                 .andExpect(status().isOk())
-                .andExpect(
-                        header().string(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + expectedArchiveName))
-                .andExpect(content().contentType("application/zip"));
+                .andExpect(content().contentType("application/zip"))
+                .andExpect(header().string("Content-Length", String.valueOf(EXPERIMENT_ACCESSION_LIST.size() * 10)))
```
alfonsomunozpomer: Why does the Content-Length header need to match that specific value? How does that indicate the validity of the zip file?

sc/src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -148,22 +154,24 @@ void downloadArchiveForInvalidExperimentAccessions() throws Exception {
...
 
-        this.mockMvc.perform(get(ARCHIVE_DOWNLOAD_LIST_URL)
-                .param("accession", EXPERIMENT_ACCESSION_LIST.get(0))
-                .param("accession", EXPERIMENT_ACCESSION_LIST.get(1))
-                .param("accession", INVALID_EXPERIMENT_ACCESSION_LIST.get(0))
-                .param("accession", INVALID_EXPERIMENT_ACCESSION_LIST.get(1)))
-                .andExpect(status().isOk())
-                .andExpect(header().string("Content-Length", String.valueOf(EXPERIMENT_ACCESSION_LIST.size() * 10)))
+        result.andExpect(status().isOk())
+                .andExpect(content().string(containsString(zipInputStream.getNextEntry().getName())))
```
alfonsomunozpomer: You should test that the zip archive contains **all** the file names you expect, not only one.

alfonsomunozpomer (changes requested):
This is a lot better. Just needs a bit of extra sugar for completeness.

alfonsomunozpomer (changes requested):
Oh no! You are now checking the raw response as a string. The fact that the name of the files is even readable is just a bit of fluke, since it’s a binary file. You should check the `ZipEntry` names against the expected file names.

alfonsomunozpomer:
Continues in https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/3.

PR #242, opened 2019-03-21
Hotfix #164800575 – Fix file leak on release metadata

alfonsomunozpomer:
I replaced the logic that reads from the JSON each time the homepage is accessed by a `LazyReference` which guarantees that the file is read only once in a thread-safe way.

The test that asks for no files is not strictly true, since the file could be left open, but wrapping it in a `LazyReference` ensures that at most one file handle is used. Testing that it’s a `LazyReference` exposes implementation details that the test shouldn’t care about, but we cannot test that a file referenced by a `Path` isn’t kept open by the JRE.

### DIFF
--- a/base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
+++ b/base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
@@ -1,33 +1,45 @@
 package uk.ac.ebi.atlas.species;
 
+import com.atlassian.util.concurrent.LazyReference;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.stream.JsonReader;
-import uk.ac.ebi.atlas.model.resource.AtlasResource;
-import uk.ac.ebi.atlas.model.resource.JsonFile;
-import uk.ac.ebi.atlas.utils.GsonProvider;
+import org.springframework.stereotype.Component;
 
-import javax.inject.Inject;
-import javax.inject.Named;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
-import java.util.Map;
 
-@Named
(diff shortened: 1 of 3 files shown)


base/src/main/java/uk/ac/ebi/atlas/species/AtlasInformationDao.java
```diff
@@ -1,33 +1,45 @@
...
+                        JsonReader jsonReader =
+                                new JsonReader(
+                                        Files.newBufferedReader(atlasInformationFilePath, StandardCharsets.UTF_8));
+                        ImmutableMap<String, String> atlasInformation =
+                                ImmutableMap.copyOf(GSON.<HashMap<String, String>>fromJson(jsonReader, HashMap.class));
+                        jsonReader.close();
+                        return atlasInformation;
+                    } catch (IOException e) {
+                        return ImmutableMap.of(
+                                "ensembl", "foo",
```
monicajianu: If the file read fails for whatever reason, then users will see "foo", right? I think something a little less software engineering-y would look better.

alfonsomunozpomer: Oh, I missed that, I wanted to have “unknown” in all fields. Will fix.

base/src/test/java/uk/ac/ebi/atlas/species/AtlasInformationDaoTest.java
```diff
@@ -0,0 +1,75 @@
...
+                        ImmutableMap.of(
+                                "ensembl", "94",
+                                "ensembl_genomes", "41",
+                                "wormbase_parasite", "11",
+                                "efo", "2.98"));
+    }
+
+    @Test
+    void noFileLeaksPlease() {
+        assertThat(subject.atlasInformation)
```
monicajianu, line 58: Hmm... creative indeed 😟

monicajianu (approved):
Looks good. Unrelated: shouldn't we be showing this kind of information in SC as well?

alfonsomunozpomer:
“foo” changed. On the unrelated note: yes, we should, and we will. It can go in the next software release.

PR #243, opened 2019-03-28
Add endpoint to get metadata for tsne plot widget

haideriqbal:
For TSne plot widget there was a need to create an endpoint which will fetch metadata. Created a new endpoint in JsonExperimentTSnePlotController.java. 
There was a case when the experiment doesn't have metadata so to handle that implemented a check which returns an empty array if there is no metadata.

Link to green builds:
http://bar:8085/browse/ATLAS-GXACI21-2
http://bar:8085/browse/ATLAS-SCXACI20-GITCO-3

### DIFF
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
@@ -59,10 +59,7 @@ public class ExperimentPageContentService {
     }
 
     public JsonArray getTsnePlotMetaData(String experimentAccession) {
-        JsonObject result = new JsonObject();
-        result.add("metadata", getMetadata(experimentAccession));
-
-        return result.getAsJsonArray("metadata");
+        return getMetadata(experimentAccession);
     }
 
     public JsonObject getExperimentDesign(String experimentAccession,
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
@@ -46,15 +46,7 @@ public class ExperimentPageContentService {
         tsnePlotSettingsService.getAvailablePerplexities(experimentAccession).forEach(perplexityArray::add);
         result.add("perplexities", perplexityArray);
 
-        JsonArray metadataArray = new JsonArray();
-        Stream.concat(
-                cellMetadataDao.getMetadataFieldNames(experimentAccession).stream(),
-                cellMetadataDao.getAdditionalAttributesFieldNames(experimentAccession).stream())
-                .map(x -> ImmutableMap.of("value", x.name(), "label", x.displayName()))
-                .collect(Collectors.toSet())
-                .forEach(x -> metadataArray.add(GSON.toJsonTree(x)));
-
-        result.add("metadata", metadataArray);
+        result.add("metadata", getMetadata(experimentAccession));
 
         JsonArray units = new JsonArray();
         units.add("TPM");
@@ -62,6 +54,13 @@ public class ExperimentPageContentService {
         result.add("units", units);
 
         result.addProperty("suggesterEndpoint", "json/suggestions");
+
+        return result;
+    }
+
+    public JsonObject getTsnePlotMetaData(String experimentAccession) {
+        JsonObject result = new JsonObject();
+        result.add("metadata", getMetadata(experimentAccession));
 
         return result;
     }
@@ -144,5 +143,17 @@ public class ExperimentPageContentService {
         section.add("files", files);
 
         return section;
+    }
+
+    private JsonArray getMetadata(String experimentAccession) {
+        JsonArray metadataArray = new JsonArray();
+        Stream.concat(
+                cellMetadataDao.getMetadataFieldNames(experimentAccession).stream(),
+                cellMetadataDao.getAdditionalAttributesFieldNames(experimentAccession).stream())
+                .map(x -> ImmutableMap.of("value", x.name(), "label", x.displayName()))
+                .collect(Collectors.toSet())
+                .forEach(x -> metadataArray.add(GSON.toJsonTree(x)));
+
+        return metadataArray;
     }
 }
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
@@ -58,11 +58,11 @@ public class ExperimentPageContentService {
         return result;
     }
 
-    public JsonObject getTsnePlotMetaData(String experimentAccession) {
+    public JsonArray getTsnePlotMetaData(String experimentAccession) {
         JsonObject result = new JsonObject();
         result.add("metadata", getMetadata(experimentAccession));
 
-        return result;
+        return result.getAsJsonArray("metadata");
     }
 
     public JsonObject getExperimentDesign(String experimentAccession,
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
@@ -2,7 +2,7 @@ package uk.ac.ebi.atlas.experimentpage;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.gson.JsonObject;
+import com.google.gson.JsonArray;
 import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -82,7 +82,7 @@ public class JsonExperimentTSnePlotController extends JsonExperimentController {
         Experiment experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
 
         // need to have this check in case there is no metadata for specified experiment
-        if(getTSnePlotMetadata(experiment.getAccession()).equalsIgnoreCase("metadata:   []")) {
+        if(getTSnePlotMetadata(experiment.getAccession()).equalsIgnoreCase("[]")) {
             return GSON.toJson(ImmutableMap.of(
                     "series",
                     new ArrayList<Map<String, Object>>()
@@ -147,7 +147,7 @@ public class JsonExperimentTSnePlotController extends JsonExperimentController {
     @ResponseBody
     public String getTSnePlotMetadata(
             @PathVariable String experimentAccession) {
-        JsonObject plotData = experimentPageContentService.getTsnePlotMetaData(experimentAccession);
+        JsonArray plotData = experimentPageContentService.getTsnePlotMetaData(experimentAccession);
         return GSON.toJson(plotData);
     }
 
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 @RestController
@@ -82,12 +83,11 @@ public class JsonExperimentTSnePlotController extends JsonExperimentController {
         Experiment experiment = experimentTrader.getExperiment(experimentAccession, accessKey);
 
         // need to have this check in case there is no metadata for specified experiment
-        if(getTSnePlotMetadata(experiment.getAccession()).equalsIgnoreCase("[]")) {
+        if(isNullOrEmpty(getTSnePlotMetadata(experiment.getAccession()))) {
             return GSON.toJson(ImmutableMap.of(
                     "series",
                     new ArrayList<Map<String, Object>>()
             ));
-
         }
         else {
             return GSON.toJson(
@@ -148,7 +148,11 @@ public class JsonExperimentTSnePlotController extends JsonExperimentController {
     public String getTSnePlotMetadata(
             @PathVariable String experimentAccession) {
         JsonArray plotData = experimentPageContentService.getTsnePlotMetaData(experimentAccession);
-        return GSON.toJson(plotData);
+        if(plotData.size() == 0) {
+            return null;
+        } else {
+            return GSON.toJson(plotData);
+        }
     }
 
     private List<Map<String, Object>> modelForHighcharts(String seriesNamePrefix, Map<?, Set<TSnePoint>> points) {
--- a/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
+++ b/sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
@@ -2,7 +2,6 @@ package uk.ac.ebi.atlas.experimentpage;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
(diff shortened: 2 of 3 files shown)


sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -125,6 +141,19 @@ public String tSnePlotWithExpression(
...
     }
 
+    @RequestMapping(
+            value = "json/experiments/{experimentAccession}/metadata",
+            method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ResponseBody
+    public String getTSnePlotMetadata(
+            @PathVariable String experimentAccession) {
+        JsonObject plotData = experimentPageContentService.getTsnePlotData(experimentAccession);
```
monicajianu: The `experimentPageContentService` is something that was specifically designed to pass exactly the right data to the experiment page jsp. If you are rewriting the t-sne plot React component to rely on endpoints only, you should also refactor the backend. Especially since you are calling a method that is actually returning a lot more than you need, and then you end up manipulating JSON objects. 
Try to move logic from the `ExperimentPageContentService` to where you need it, and deal with Java objects as much as possible. Turning something to JSON should be the last step the controller does. Also, if some logic is no longer required for the React component, then don't hesitate to remove it!

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
```diff
@@ -22,22 +24,19 @@
...
 
-import static org.hamcrest.Matchers.everyItem;
-import static org.hamcrest.Matchers.greaterThan;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.isA;
-import static org.hamcrest.Matchers.oneOf;
+import java.util.Random;
+
+import static org.hamcrest.Matchers.*;
```
monicajianu: We try to add every import, as opposed to using `.*`. This might be IntelliJ trying to be clever, you might need to change your settings:

Screenshot 2019-04-01 at 10 56 39

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
```diff
@@ -164,4 +163,29 @@ void validJsonForInvalidK() throws Exception {
                 .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
                 .andExpect(jsonPath("$.series", hasSize(1)));
     }
+
+    @Test
+    void validJsonWithValidExperimentAccession() throws Exception {
+        String [] experimentAccession = {"E-MTAB-5061", "E-EHCA-2"};
```
monicajianu, line 173: Try to use either `jdbcTestUtils` to fetch random accessions, or use parameterised tests to exhaustively test experiments (see for example `CellMetadataServiceIT`).

monicajianu (changes requested):
Good start, but I think the code logic could be organised a little better!

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
haideriqbal: I was using that approach first but issues arise when it comes to experiment with no metadata. So have to hard code to experiments that have metadata to test.

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
haideriqbal: Updated my settings. Thanks!

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: I have refactored the code.
Now I've created a separate function in `ExperimentPageContentService` which just returns TsnePlotMetadata.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
```diff
@@ -66,6 +58,13 @@ public JsonObject getTsnePlotData(String experimentAccession) {
         return result;
     }
 
+    public JsonArray getTsnePlotMetaData(String experimentAccession) {
```
monicajianu, line 61: What is this method actually doing? Why do you need to a create an JSON object, add something to it, then retrieve it immediately afterwards? What is the purpose of this intermediate step?

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -76,14 +81,24 @@ public String tSnePlotWithMetadata(
...
-        return GSON.toJson(
-                ImmutableMap.of(
-                        "series",
-                        modelForHighcharts(
-                                "",
-                                new TreeMap<>(
-                                        tSnePlotService.fetchTSnePlotWithMetadata(
-                                                experiment.getAccession(), perplexity, metadata)))));
+        // need to have this check in case there is no metadata for specified experiment
+        if(getTSnePlotMetadata(experiment.getAccession()).equalsIgnoreCase("[]")) {
```
monicajianu: Is returning an empty series a requirement for the widget? In the current implementation, with the [added support for missing metadata](https://github.com/ebi-gene-expression-group/atlas/pull/244), you get the t-SNE points which belong to a "Not available" series. 

If you did want to implement some support for requesting non-existent metadata types, I think making use of an appropriate HTTP error code might be better and more informative for the user.

monicajianu: Also, when interacting with `JsonArray`s or `JsonObject`s, try to avoid as much as possible doing any kind of string manipulations/comparisons. Try to use methods provided by the class to check if array is empty (for example, `size()`). Alternatively, pass Java objects arround, such as Lists or Maps, and interact with them instead. It will be much less prone to errors.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
haideriqbal: Agreed. It was a useless intermediate step. Fixed the logic.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: This check was added as a safety net to display a scatter plot with hard coded "k" value in the tsne widget in case there is no metadata. I'm not sure if that can be achieved with the current implementation. I have not tested it yet. 
I guess showing HTTP error codes to the user might not be a great idea with respect to UX design. There is a link to atlas experiment in the footer of the widget which users can refer to if they require more info.

Regarding string manipulation. I have fixed the logic. Now using JsonArray size function

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/ExperimentPageContentService.java
monicajianu: Why do you need the logic to be in a private method anyway? That would only be useful if you are calling the `getMetadata` in multiple different places, which you are not doing.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
monicajianu: When you are making an API endpoint, you have to think of the "UX" of the API. If I am using a REST API, if I get a 404 Not Found error, I immediately know what that means. It is up to the client (in this case the t-SNE widget, I assume), to handle the error code appropriately. If the client handles it badly and reveals the HTTP code to the end user, then yes, it would be bad UX.

I'm not sure what you mean by a hard coded k value. This endpoint is retrieving t-SNE points grouped by metadata, there shouldn't be any k values involved.

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
monicajianu: You could make use of parameterized tests and provider methods, see an example of this here: https://github.com/ebi-gene-expression-group/atlas/blob/dev/sc/src/test/java/uk/ac/ebi/atlas/metadata/JsonCellMetadataControllerWIT.java#L111

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -125,6 +140,21 @@ public String tSnePlotWithExpression(
...
+    @RequestMapping(
+            value = "json/experiments/{experimentAccession}/metadata",
+            method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ResponseBody
+    public String getTSnePlotMetadata(
+            @PathVariable String experimentAccession) {
+        JsonArray plotData = experimentPageContentService.getTsnePlotMetaData(experimentAccession);
+        if(plotData.size() == 0) {
+            return null;
```
monicajianu, line 152: Adding this has caused your test to fail. Can you think why?

monicajianu: Also, what is your reasoning for not letting Gson handle empty results?

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -76,14 +82,23 @@ public String tSnePlotWithMetadata(
...
-        return GSON.toJson(
-                ImmutableMap.of(
-                        "series",
-                        modelForHighcharts(
-                                "",
-                                new TreeMap<>(
-                                        tSnePlotService.fetchTSnePlotWithMetadata(
-                                                experiment.getAccession(), perplexity, metadata)))));
+        // need to have this check in case there is no metadata for specified experiment
+        if(isNullOrEmpty(getTSnePlotMetadata(experiment.getAccession()))) {
```
monicajianu, line 86: If the result of `getTsnePlotMetadata` is a `JsonArray`, why are you using `isNullOrEmpty` here, and `.size() == 0` later on? You have to be consistent in your implementation, especially in the same class when the discrepancy is so visible!

monicajianu (changes requested):
A few more things to address. Also, please add some tests for the `/json/experiments/{experimentAccession}/tsneplot/{perplexity}/metadata/{metadata}` endpoint.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
```diff
@@ -125,6 +140,21 @@ public String tSnePlotWithExpression(
         return GSON.toJson(model);
     }
 
+    @RequestMapping(
+            value = "json/experiments/{experimentAccession}/metadata",
+            method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ResponseBody
```
monicajianu, line 147: You don't need the `@ResponseBody` annotation here (read about it here: https://www.baeldung.com/spring-request-response-body)

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: Can you please let me know which test case is failing?

In the case of empty result `GSON.toJson(plotData)` returns `[]` as a String which I don't require and to handle it in my empty metadata check I have to do a direct string manupilation. To avoid that I return null in case of an empty array.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: The result of getTsnePlotMetadata is a String so that's why I have to treat return result as String to check if it is empty or null.

Regarding `.size()==0` here I'm dealing with JsonArray so depending on if it is empty or not I am returning null or actual value respectively.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: Thanks! didn't know if we already have annotated the controller as `@RestController` then we don't need to have `@ResponseBody`

Will update the code.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
haideriqbal: The endpoint is actually being consumed by [scxa-experiment-tsne-plot-view](https://github.com/ebi-gene-expression-group/scxa-experiment-page-tsne-plot-view). It uses it to fill `cellClusterData` series. The idea to return empty series from backend was not to break the functionality in `scxa-experiment-page-tsne-plot-view` as it throws [error](https://github.com/ebi-gene-expression-group/scxa-experiment-page-tsne-plot-view/blob/d79deb4bf2b61e3527af29af26f36c70beac2bcb/src/TSnePlotView.js#L109) if there is a bad response.

Regarding hardcoded 'k-value', in 'tsne-widget' we only show one dropdown with just metadata as options so if there is no metadata in an experiment then we still need to show scatter plot but in that case, we have to provide some `k-value`.

sc/src/test/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotControllerWIT.java
haideriqbal: I wasn't aware of this functionality to filter accessions. Will fix and update code.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
monicajianu: Okay, in that case, when would the t-SNE plot ever request non-existent metadata, seeing as it is retrieving the metadata from the backend in the first place?

As for the tsne widget, I assume it is doing a separate request to retrieve data grouped by k if no metadata metadata exists, i.e. if the request to `json/experiments/{experimentAccession}/metadata` returns empty results, then retrieve tsne points grouped by k (by doing a separate request). The endpoints shouldn't be responsible for making decisions about empty data, the client should decide what to do about it.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
monicajianu: Have a look at [Bamboo](http://bar:8085/browse/ATLASOLD-SCXACI20-GITCO-1/test/case/57658993) for the failing test.

sc/src/main/java/uk/ac/ebi/atlas/experimentpage/JsonExperimentTSnePlotController.java
monicajianu: Sorry about that, I didn't notice it was a different `getTsnePlotMetadata` (the service has the same method name so I think I got confused). I think the logic can still be simplified, and this check might not even be needed (see my other comment).

alfonsomunozpomer:
This PR continues in https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/6.

PR #244, opened 2019-04-04
Bugfix #164934771 - Handle blank factor or characterstics

monicajianu:
Finally finished this! Added a pretty naive fix for missing factors/characteristics, by replacing them with the string "not available" at DAO level.

I also found a new data production bug with the new experiment I added, we have some missing files, which is why one experiment is failing in Bamboo.


sc/src/test/java/uk/ac/ebi/atlas/metadata/CellMetadataDaoIT.java
```diff
@@ -147,6 +150,24 @@ void invalidCellIdAndExperimentAccessionHasNoFactorFields() {
...
 
+    @Test
+    void experimentWithMissingValuesReturnsNotAvailable() {
+        String experimentAccession = "E-GEOD-71585";
+
+        ImmutableList<String> cellIdsWithMissingValues = ImmutableList.of(
+                "SRR2138737",
+                "SRR2140225",
+                "SRR2139550",
+                "SRR2139566");
```
alfonsomunozpomer, line 163: As a future improvement these could be randomly sampled from Solr. Could you add a TODO comment?

sc/src/test/java/uk/ac/ebi/atlas/metadata/CellMetadataDaoIT.java
```diff
@@ -147,6 +150,24 @@ void invalidCellIdAndExperimentAccessionHasNoFactorFields() {
...
+
+        Map<String, String> result = subject.getMetadataValueForCellIds(experimentAccession,
+                SingleCellAnalyticsCollectionProxy.CHARACTERISTIC_INFERRED_CELL_TYPE,
+                cellIdsWithMissingValues);
+
+        assertThat(result.values()).containsOnly("not available");
```
alfonsomunozpomer: Rather than hard-coding the string in the class and the test, I think it’s better to have the tested class have a `static final String` variable that can be checked by the test class.

sc/src/test/java/uk/ac/ebi/atlas/search/GeneSearchServiceIT.java
```diff
@@ -60,7 +67,10 @@ void cleanDatabaseTables() {
     @MethodSource("experimentAccesionWithoutPreferredKProvider")
     void experimentsWithoutPreferredKReturnASingleProfile(String experimentAccession) {
         String geneId = jdbcTestUtils.fetchRandomMarkerGeneFromSingleCellExperiment(experimentAccession);
-        assertThat(subject.getMarkerGeneProfile(geneId)).hasSize(1);
+        LOGGER.info("Fetching marker gene profile for {}", geneId);
+        Map<String, Map<String, Map<Integer, List<Integer>>>> markerGeneProfile = subject.getMarkerGeneProfile(geneId);
+
+        assertThat(markerGeneProfile).hasSize(1);
     }
```
alfonsomunozpomer, line 74: [Is this test passing](https://www.pivotaltracker.com/story/show/164880470)?

alfonsomunozpomer (approved):
Very minor changes. Also, see my question about the test that verifies that experiments without a preferred *k* return a single marker gene profile. It falls outside the scope of this fix, but it makes me wonder...

